PR #48, opened 2023-09-25
fix: Fix merkle-tree related bug in tracing & ST Prover

rpanic:


packages/sequencer/test/integration/BlockProduction.test.ts
```diff
@@ -298,7 +301,47 @@ describe("block production", () => {
     );
   }, 160_000);
 
-  it.only("should produce block with a tx with a lot of STs", async () => {
+  it.only.each([
```
maht0rz, line 304: `.only` should be removed i think

maht0rz (changes requested):

packages/sequencer/test/integration/BlockProduction.test.ts
maht0rz: agreed to leave .only here since the tests are inherently broken and need to be fixed anyways

maht0rz (approved):

PR #50, opened 2023-09-29
[WIP] feat: added struct as param for transactions

maht0rz:
This one works, but it's typed pretty poorly, i want to fix it before we merge.

### DIFF
--- a/packages/sdk/src/appChain/AppChain.ts
+++ b/packages/sdk/src/appChain/AppChain.ts
@@ -28,7 +28,7 @@ import {
   ProtocolModule,
 } from "@proto-kit/protocol";
 import { container } from "tsyringe";
-import { Field, PublicKey, UInt64 } from "snarkyjs";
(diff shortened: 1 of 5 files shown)


packages/sdk/src/appChain/AppChain.ts
```diff
@@ -207,13 +207,31 @@ export class AppChain<
       );
     }
 
-    const argsFields = args.flatMap((arg) => arg.toFields());
+    // forgive me, i'll fix this type issue soon
+    const runtimeModule = this.runtime.resolve(moduleName as any);
```
maht0rz, line 211: @rpanic if i try to remove this `any` using `this.runtime.assertIsValidModuleName`, i get the following error:

```
Assertions require every name in the call target to be declared with an explicit type annotation. ts(2775)
```

Haven't yet figured out how to solve it

rpanic: Then I'd say leave it for now

rpanic (approved):

PR #52, opened 2023-10-03
Tiered DI

rpanic:


framework.iml
```diff
@@ -2,7 +2,9 @@
 <module type="WEB_MODULE" version="4">
   <component name="NewModuleRootManager" inherit-compiler-output="true">
     <exclude-output />
-    <content url="file://$MODULE_DIR$" />
+    <content url="file://$MODULE_DIR$">
```
maht0rz, line 5: i thought we're gitignoring IDE specific files no? we should also remove `.vscode` if yes

packages/common/src/config/ChildContainerStartable.ts
```diff
@@ -0,0 +1,5 @@
+import { ChildContainerProvider } from "./ChildContainerProvider";
+
+export interface ChildContainerStartable {
```
maht0rz: probably could have been `Creatable` not `Startable`

maht0rz: I wish we renamed `config/` to `moduleContainer/` but its not in the scope here.

packages/common/src/config/ConfigurableModule.ts
```diff
@@ -36,6 +43,11 @@ export class ConfigurableModule<Config> implements Configurable<Config> {
   public set config(config: Config) {
     this.currentConfig = config;
   }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  public create(childContainerProvider: ChildContainerProvider): void {
+    noop();
```
maht0rz, line 49: Disabling `@typescript-eslint/no-empty-function` for this line would be nicer than `noop()`

packages/common/src/config/ModuleContainer.ts
```diff
@@ -40,7 +48,9 @@ const errors = {
 export const ModuleContainerErrors = errors;
 
 // determines that a module should be configurable by default
-export type BaseModuleType = TypedClass<Configurable<unknown>>;
+export type BaseModuleType = TypedClass<
+  ChildContainerStartable & Configurable<unknown>
```
maht0rz: Would be nice if we did a merged interface for this joint type and used it in `ConfigurableModule implements ...` as well

packages/common/src/config/ModuleContainer.ts
```diff
@@ -148,6 +161,16 @@ export class ModuleContainer<
     }
   }
 
+  public assertContainerInitialized(
```
maht0rz, line 176: We should probably start avoiding in-class assertion methods since we've learned recently they dont work outside of classes/instances themselves. But out of scope here.

packages/common/src/config/ModuleContainer.ts
```diff
@@ -148,6 +161,16 @@ export class ModuleContainer<
...
+    container: DependencyContainer | undefined
+  ): asserts container is DependencyContainer {
+    if (container === undefined) {
+      throw new Error(
+        "DependencyContainer not set. Be sure to only call DI-related function in create() and not inside the constructor."
```
maht0rz: Can we extract the error?

maht0rz: Would be nice if we could add `this.constructor.name` to the error, assuming it'll print the class that inherited the `ModuleContainer`

packages/common/src/config/ModuleContainer.ts
```diff
@@ -162,9 +185,11 @@ export class ModuleContainer<
 
         log.debug(`Registering module: ${moduleName}`);
 
+        const definitionEntry = modules[moduleName];
+
         this.container.register(
           moduleName,
-          { useClass: modules[moduleName] },
+          { useClass: definitionEntry },
```
maht0rz, line 202: By extracting this into a variable it throws a shorthand properties eslint error.

packages/common/src/config/ModuleContainer.ts
```diff
@@ -234,7 +269,10 @@ export class ModuleContainer<
     const isValidModuleInstance = instance instanceof moduleType;
 
     if (!isValidModuleInstance) {
-      throw new Error("Incompatible module instance");
+      console.log(instance);
```
maht0rz: 👀

packages/common/src/config/ModuleContainer.ts
```diff
@@ -234,7 +269,10 @@ export class ModuleContainer<
...
+      throw new Error(
+        `Incompatible module instance ("${moduleName}" not instanceof ${moduleType.name})`
```
maht0rz: Can we extract this error please?

packages/common/src/config/ModuleContainer.ts
```diff
@@ -271,8 +309,30 @@ export class ModuleContainer<
           throw errors.unableToDecorateModule(containedModuleName);
         }
         this.decorateModule(moduleName, containedModule);
+
+        containedModule.create(() => {
+          const container = this.container.createChildContainer();
+          container.reset();
```
maht0rz, line 322: AFAIK this resets any registered dependencies from the parent. I thought the purpose of passing child containers from the parent to the child module, is to preserve access to parent deps.

packages/common/src/config/ModuleContainer.ts
```diff
@@ -271,8 +309,30 @@ export class ModuleContainer<
...
     );
   }
+
+  /**
+   * This is a placeholder for individual modules to override.
+   * This method will be called whenever the underlying container fully
+   * initialized
+   */
+  public create(childContainerProvider: ChildContainerProvider): void {
+    this.providedContainer = childContainerProvider();
```
maht0rz, line 336: Following up the thought train from line #315, shouldn't `childContainerProvider()` be called after `registerModule` so that the child container has access to parent deps? This is probably not how it works under the hood and the order here doesnt matter since the parent container instance remains the same, it just semantics.

packages/common/src/config/ModuleContainer.ts
```diff
@@ -271,8 +309,30 @@ export class ModuleContainer<
...
+
+    // register all provided modules when the container is created
+    this.registerModules(this.definition.modules);
+
+    this.registerValue({
+      ChildContainerProvider: () => this.container.createChildContainer(),
```
maht0rz, line 342: Wouldn't it be nicer if the injection token was `childContainerProvider` since its just a function, and not a class? But i get the follow up on the original interface name `ChildContainerProvider`

packages/common/src/utils.ts
```diff
@@ -35,6 +35,9 @@ export function dummyValue<Value>(
   return valueType.fromFields(fields) as Value;
 }
 
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+export function noop(): void {}
```
maht0rz, line 39: could have used `lodash/noop` 😂

packages/module/src/runtime/Runtime.ts
```diff
@@ -190,12 +191,18 @@ export class RuntimeZkProgrammable<
 @injectable()
 export class Runtime<Modules extends RuntimeModulesRecord>
   extends ModuleContainer<Modules>
-  implements WithZkProgrammable<undefined, MethodPublicOutput>
+  implements
+    WithZkProgrammable<undefined, MethodPublicOutput>,
```
maht0rz: This is already a part of `RuntimeEnvironment`

packages/module/src/runtime/Runtime.ts
```diff
@@ -223,15 +230,16 @@ export class Runtime<Modules extends RuntimeModulesRecord>
...
   // eslint-disable-next-line no-warning-comments
   // TODO Remove after changing DFs to type-based approach
-  public start() {
-    this.registerValue({
-      Runtime: this,
-    });
+  public create(childContainerProvider: ChildContainerProvider) {
+    super.create(childContainerProvider);
+
+    console.log("Registering create()");
```
maht0rz: 👀

packages/module/test/modules/State.test.ts
```diff
@@ -39,7 +39,7 @@ describe("transient state", () => {
       },
     });
 
-    runtime.dependencyContainer.register("AppChain", {
+    runtime.dependencyContainer.register("AreProofsEnabled", {
```
maht0rz, line 42: Shouldn't runtime be using `runtime.providedContainer` which is provided from the AppChain? Or since Runtime isnt an AppChain module it can't receive a childContainerFactory?

packages/protocol/src/protocol/Protocol.ts
```diff
@@ -85,32 +75,6 @@ export class Protocol<
...
-
-    // We need this so that tsyringe doesn't throw when no hooks are registered
-    if (!atLeastOneTransactionHookRegistered) {
-      this.container.register(
-        "ProvableTransactionHook",
-        { useClass: NoopTransactionHook },
-        { lifecycle: Lifecycle.ContainerScoped }
-      );
-    }
-    // this.container.afterResolution<ProvableTransactionHook>("ProvableTransactionHook", (token, result) => {
```
maht0rz, line 111: 👀

packages/protocol/src/protocol/Protocol.ts
```diff
@@ -162,6 +122,52 @@ export class Protocol<
...
+        "ProvableTransactionHook",
+        { useClass: NoopTransactionHook },
+        { lifecycle: Lifecycle.ContainerScoped }
+      );
+    }
+
+    // Set empty config for all modules, since we don't have that feature yet
+    // eslint-disable-next-line max-len
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any,@typescript-eslint/no-unsafe-assignment
+    const emptyConfig = Object.keys(this.definition.modules).reduce<any>(
```
maht0rz: if we want to use this, it should be part of the `super.create(...)` so that every component can use it equally (like runtime)

packages/protocol/src/protocol/ProtocolModule.ts
```diff
@@ -1,20 +1,26 @@
...
 
   public get appChain(): AreProofsEnabled | undefined {
-    return this.protocol?.dependencyContainer.resolve<AreProofsEnabled>(
-      "AppChain"
-    );
+    return this.protocol?.getAreProofsEnabled();
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  public create(childContainerProvider: ChildContainerProvider): void {
```
maht0rz, line 20: should ProtocolModule extend ConfigurableModule and therefore get this function for free?

packages/sdk/src/appChain/AppChain.ts
```diff
@@ -252,43 +323,27 @@ export class AppChain<
...
   public async start() {
-    [this.runtime, this.protocol, this.sequencer].forEach((container) => {
-      container.registerValue({ AppChain: this });
-    });
+    super.create(() => container);
 
-    this.protocol.registerValue({
-      Runtime: this.runtime,
-    });
+    // These three statements are crucial for dependencies inside any of these
```
maht0rz, line 328: can this be a multiline comment instead?

packages/sdk/src/appChain/AreProofsEnabledFactory.ts
```diff
@@ -0,0 +1,28 @@
+import { injectable } from "tsyringe";
+import {
+  AreProofsEnabled,
+  dependency,
+  dependencyFactory,
+  DependencyFactory,
+} from "@proto-kit/common";
+
+@injectable()
+export class AreProofsEnabledImpl implements AreProofsEnabled {
```
maht0rz: -> `InMemoryAreProofsEnabled` would be nicer, `-Impl` or `I-` for interfaces is something we should avoid right

packages/sdk/src/appChain/TestingAppChain.ts
```diff
@@ -110,12 +111,12 @@ export class TestingAppChain<
       QueryTransportModule: {},
     });
   }
-
-  public useAuroSigner() {
-    this.registerModules({
-      Signer: AuroSigner,
-    });
-  }
+  //
```
maht0rz, line 115: i will fix this once its merged to develop, before publishing the next packages.

packages/sequencer/src/worker/queue/LocalTaskQueue.ts
```diff
@@ -31,7 +35,9 @@ export class LocalTaskQueue implements TaskQueue {
     [key: string]: QueueListener[];
   } = {};
 
-  public constructor(private readonly simulatedDuration?: number) {}
+  public constructor() {
```
maht0rz: probably can be removed all together

packages/sequencer/test/integration/BlockProduction.test.ts
```diff
@@ -253,7 +261,7 @@ describe("block production", () => {
 
   const numberTxs = 3;
 
-  it("should produce block with multiple transaction", async () => {
+  it.only("should produce block with multiple transaction", async () => {
```
maht0rz, line 264: 👀

maht0rz (changes requested):

packages/common/src/config/ConfigurableModule.ts
rpanic: Ok but then why do we have this rule when we don't want the thing that the rule forces us to do?

framework.iml
rpanic: I don't know tbh, I have seen most project I worked in commit these kind of files to make IDE setup easier. And since I apparently committed in some time in the past, I though to just include it

packages/common/src/config/ModuleContainer.ts
rpanic: Can we disable this rule? In my opinion it actually reduces readability in a lot of cases. Bcs renaming it useClass would remove information

packages/common/src/config/ModuleContainer.ts
rpanic: that's your code 😂

packages/common/src/config/ModuleContainer.ts
rpanic: resolved in discord

packages/common/src/config/ModuleContainer.ts
rpanic: I think it has more value to start our injection tokens consistently with a Uppercase letter. I know it is a function but we don't distinguish between services and primitives (yet). Let me know if you think differently

packages/module/src/runtime/Runtime.ts
rpanic: good catch!

packages/module/test/modules/State.test.ts
rpanic: It does, under the hood, `.dependencyContainer` returns `providedContainer` but does an additional assertion that it is set to throw a nice error in the case it is not

packages/protocol/src/protocol/ProtocolModule.ts
rpanic: I made the ProtocolModules Configurable now

packages/sdk/src/appChain/AppChain.ts
rpanic: no there's a eslint rule against that

packages/sequencer/test/integration/BlockProduction.test.ts
rpanic: explained that in a previous PR, we can't execute multiple tests at once rn. 
Maybe that has changed now though?

PR #53, opened 2023-10-04
Fix methodId encoding

maht0rz:


packages/sdk/src/appChain/AppChain.ts
```diff
@@ -207,13 +207,31 @@ export class AppChain<
       );
     }
 
-    const argsFields = args.flatMap((arg) => arg.toFields());
+    // forgive me, i'll fix this type issue soon
+    const runtimeModule = this.runtime.resolve(moduleName as any);
```
rpanic, line 211: If you want to fix that, add `this.runtime.assertIsValidModuleName(this.definition.modules, moduleName);` before that. But you might also have to do this https://github.com/microsoft/TypeScript/issues/36931#issuecomment-846131999

rpanic (approved):

packages/sdk/src/appChain/AppChain.ts
rpanic: ah this should actually be in #50

PR #54, opened 2023-10-04
Fix tracing bug

### LINKED ISSUE #51: Multiple transactions in single block errors with `MerkleWitness not valid for StateTransition`
**Describe the bug**
When sending many transactions to an app-chain & then producing a block the test errors with:
```bash
test/Balances.test.ts:
Block Production time: 1696 milliseconds
111 |         if (response.taskId !== undefined) {
112 |             const resolveFunction = this.resultsPending[response.taskId];
113 |             if (!this.erroredOut) {
114 |                 if (response.status === "error") {
115 |                     this.erroredOut = true;
116 |                     this.errorFunction?.(new Error(`Error in worker: ${response.payload}`));
                                            ^
error: Error in worker: MerkleWitness not valid for StateTransition (0, type normal)
Bool.assertTrue(): false != true
```

I came across this error in both `bun` and `npm`.

**To Reproduce**
1. Go to [bun bug repository](https://github.com/teddyjfpender/proto-kit-bun-project) or the [npm bug repository](https://github.com/teddyjfpender/proto-kit-npm-project)
2. Run the tests (`bun install && bun test` or `npm test`)
3. See error

**Expected behavior**
When sending multiple transactions to the app-chain it should produce a block and not error.

**Additional context**

rpanic:
Fix bug where previously written merkle-tree entries got overwritten by preloading

closes #51

### DIFF
--- a/packages/protocol/src/prover/statetransition/StateTransitionProver.ts
+++ b/packages/protocol/src/prover/statetransition/StateTransitionProver.ts
@@ -203,6 +203,12 @@ export class StateTransitionProverProgrammable extends ZkProgrammable<
       transition.from.value
     );
 
+    // if (!membershipValid.toBoolean()) {
+    //   console.log(
+    //     `Merklewitness error ${root1.toString()} ${root2.toString()}`
+    //   );
+    // }
+
     membershipValid
       .or(transition.from.isSome.not())
       .assertTrue(
--- a/packages/protocol/src/prover/statetransition/StateTransitionProver.ts
+++ b/packages/protocol/src/prover/statetransition/StateTransitionProver.ts
@@ -203,12 +203,6 @@ export class StateTransitionProverProgrammable extends ZkProgrammable<
       transition.from.value
     );
 
-    // if (!membershipValid.toBoolean()) {
-    //   console.log(
-    //     `Merklewitness error ${root1.toString()} ${root2.toString()}`
-    //   );
-    // }
-
     membershipValid
       .or(transition.from.isSome.not())
       .assertTrue(
--- a/packages/sequencer/src/protocol/production/TransactionTraceService.ts
+++ b/packages/sequencer/src/protocol/production/TransactionTraceService.ts
@@ -278,6 +278,8 @@ export class TransactionTraceService {
     let stateRoot = initialRoot;
     let protocolStateRoot = initialRoot;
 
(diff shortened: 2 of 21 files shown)


packages/protocol/src/prover/statetransition/StateTransitionProver.ts
```diff
@@ -203,6 +203,12 @@ export class StateTransitionProverProgrammable extends ZkProgrammable<
       transition.from.value
     );
 
+    // if (!membershipValid.toBoolean()) {
```
maht0rz: seems like an artifact from debuging, should we remove this?

packages/sequencer/src/protocol/production/TransactionTraceService.ts
```diff
@@ -397,35 +485,44 @@ export class TransactionTraceService {
...
+  ): Promise<{
     runtimeKeys: Field[];
     protocolKeys: Field[];
-  } {
+  }> {
     // Execute the first time with dummy service
-    this.runtime.stateServiceProvider.setCurrentStateService(
-      this.dummyStateService
-    );
+    // this.runtime.stateServiceProvider.setCurrentStateService(
```
maht0rz: probably a debug artifact, should we remove?

packages/sequencer/src/protocol/production/execution/CachedStateService.ts
```diff
@@ -32,13 +35,15 @@ export class CachedStateService
...
     if (this.parent !== undefined && this.get(key) === undefined) {
       const value = await this.parent.getAsync(key);
-      log.debug(
-        `Preloading ${key.toString()}: ${
-          // eslint-disable-next-line max-len
-          // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
-          value?.map((element) => element.toString()) ?? []
-        }`
-      );
+      if (this.debugLogs) {
```
maht0rz: do we need this extra layer of logging enable/disable? wouldnt setLevel('DEBUG') be enought?

rpanic: I think so, yes. This i mainly for debugging issues on our side, so nothing that should be logged for the user. And since there is no log level belug DEBUG, i opted for this approach

maht0rz (approved):

PR #58, opened 2023-10-12
feat: changed tx args transport format to JSON + proof support

maht0rz:
This PR changes the args transport layer to JSON[] from Field[], fixing encoding/decoding of certain args types such as Struct or Signature. 

Proof support will be also possible thanks to this PR - ~but we need to add proof argsHash support using public input/output in a separate PR~


packages/common/src/utils.ts
```diff
@@ -45,6 +45,19 @@ export interface ToFieldable {
...
 
+export interface ToFieldableStatic {
+  toFields: (value: unknown) => Field[];
+}
+
+export interface ToJSONableStatic {
+  toJSON: (value: unknown) => any;
+}
+
+export interface ProofTypes {
```
rpanic, line 56: Can we rename this to something more verbose?

packages/api/src/graphql/modules/QueryGraphqlModule.ts
```diff
@@ -150,13 +149,12 @@ export class QueryGraphqlModule<
   ): GraphQLScalarType | ObjectType {
     // This is a temporary workaround until transport-layer has been
```
rpanic, line 191: I think I wanted this to be changed somehow when doing the json transport layer - but I am not sure anymore how exactly I wanted this to change. So I guess we can remove this comment 🤔

packages/api/src/graphql/modules/QueryGraphqlModule.ts
```diff
@@ -150,13 +149,12 @@ export class QueryGraphqlModule<
...
     // switched to json
-    const valueType = type as FlexibleProvablePure<ProvableType> &
-      FromFieldClass &
-      ProvableExtension<any>;
+    const valueType = type as ProvableExtended<unknown>;
     const valueFieldLength = MethodParameterDecoder.fieldSize(valueType);
 
     const dummyValue = valueType.fromFields(
-      range(0, valueFieldLength).map(() => Field(0))
+      range(0, valueFieldLength).map(() => Field(0)),
```
rpanic, line 197: Shouldn't we use `ProvablePureExtended` so we can skip the auxilary data input here? Since the input type is of `FlexibleProvablePure`

packages/common/src/utils.ts
```diff
@@ -45,6 +45,19 @@ export interface ToFieldable {
   toFields: () => Field[];
 }
 
+export interface ToFieldableStatic {
```
rpanic, line 48: What does "static" in this context mean? If it is of the sense of being only used on class-types that implement the interface on a static level, we might add `new(..args: unknown[]) => unknown` to this signature so that it can't be used on objects

packages/module/src/method/MethodParameterDecoder.ts
```diff
@@ -30,12 +12,12 @@ const errors = {
 
 export class MethodParameterDecoder {
   public static fromMethod(target: RuntimeModule<unknown>, methodName: string) {
-    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
-    const paramtypes = Reflect.getMetadata(
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    const paramtypes: ProvableExtended<unknown>[] = Reflect.getMetadata(
```
rpanic, line 16: We should reflect the possibility of it being `undefined` in the type

packages/module/src/method/runtimeMethod.ts
```diff
@@ -99,19 +111,37 @@ export function toWrappedMethod(
       "Runtimemethod called with wrong methodId on the transaction object"
     );
 
-    const paramTypes: FlexibleProvable<unknown>[] = Reflect.getMetadata(
-      "design:paramtypes",
-      this,
-      methodName
-    );
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    const parameterTypes: ProofTypes[] | ToFieldableStatic[] =
```
rpanic, line 115: Either use `ArgumentTypes[]` here if applicable or `(ProofTypes | ToFieldableStatic)[]`

packages/module/src/runtime/Runtime.ts
```diff
@@ -222,8 +222,6 @@ export class Runtime<Modules extends RuntimeModulesRecord>
   // TODO Remove after changing DFs to type-based approach
   public create(childContainerProvider: ChildContainerProvider) {
     super.create(childContainerProvider);
-
-    this.registerDependencyFactories([MethodIdFactory]);
```
rpanic: Have you tested that this works? I am worried that when we move this to the AppChain, we get some dependency cycles though Runtime not being configured yet and MethodIdFactory needing an initialized runtime. I had this once during testing that. 
Alternative would be register it both on Runtime & Sequencer

packages/sdk/src/appChain/AppChain.ts
```diff
@@ -279,10 +284,15 @@ export class AppChain<
     }
 
     // forgive me, i'll fix this type issue soon
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const runtimeModule = this.runtime.resolve(moduleName as any);
 
     // find types of args for the runtime method thats being called
-    const parameterTypes: FlexibleProvable<unknown>[] = Reflect.getMetadata(
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    const parameterTypes:
```
rpanic, line 287: Again, integrate the types inside the Array `(A | B | C)[]`

packages/sequencer/src/mempool/verification/TransactionValidator.ts
```diff
@@ -37,11 +37,6 @@ export class TransactionValidator {
       methodPath[1]
     );
```
rpanic, line 37: I guess validation will be a seperate PR? Would be nice to have for IST

PR #63, opened 2023-10-19
feat: upgrade from snarkyjs to o1js

maht0rz:
This branch is a result of find and replace "snarkyjs" with "o1js" plus changing to v0.13.1. It builds locally


rpanic (approved):
lgtm

PR #69, opened 2023-10-27
Added graphql client modules

rpanic:


packages/sdk/test/graphql.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    // const server = await startServer();
+    //
+    // await sleep(2000);
+
+    appChain = prepare();
+
+    await appChain.start();
+  })
+
+  it.skip("should retrieve state", async () => {
```
maht0rz: I'd like to see a test where we submit a tx, produce a block and then read the state, checking the tx has been successfull too

packages/sdk/test/graphql.test.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  return appChain
+}
+
+describe("graphql client test", function() {
+
+  let appChain: ReturnType<typeof prepare> | undefined = undefined
+
+  beforeAll(async () => {
+    // const server = await startServer();
```
maht0rz: looks like this isnt used, so the helper fn for sleep can be removed too, or moved to common utils

packages/sdk/src/graphql.ts
maht0rz, line 1: probably should be removed in favour of test/graphql.test.ts

packages/sdk/test/graphql.test.ts
```diff
@@ -0,0 +1,158 @@
...
+  }).sign(pk);
+
+  console.log(tx.toJSON());
+
+  return tx;
+}
+
+function prepare(){
+
+  const appChain = AppChain.from({
```
maht0rz: would be great if we had to separate AppChains running, one for the server side and one for the client side, then this test would be perfect

maht0rz (changes requested):

packages/sdk/test/graphql.test.ts
rpanic: we do

PR #79, opened 2023-12-01
Fix config inheritance

### LINKED ISSUE #21: Enable ConfigurationReceivers to receive "void" as Type argument
ConfigurationReceivers have to provide `{}` instead of `void` as a type argument for the module-decorators to accept the class.

Ideally we want to be able to provide something like void or undefined for Modules that don't want to receive a configuration

### LINKED ISSUE #62: consecutive `setSigner(...)` call doesn't work, due to config not being replacable after initial setup

rpanic:
Closes #21 
Closes #62


packages/api/src/graphql/modules/MempoolResolver.ts
```diff
@@ -72,7 +73,7 @@ export class TransactionObject {
 }
 
 @graphqlModule()
-export class MempoolResolver extends GraphqlModule<object> {
+export class MempoolResolver extends GraphqlModule<NoConfig> {
```
maht0rz: why not provide a default value for the generic instead?

packages/common/src/config/ConfigurableModule.ts
```diff
@@ -33,6 +35,7 @@ export class ConfigurableModule<Config>
 
   // retrieve the existing config
   public get config(): Config {
+    console.log("C", this.currentConfig);
```
maht0rz: ew

packages/common/src/config/ModuleContainer.ts
```diff
@@ -250,13 +264,24 @@ export class ModuleContainer<
...
+      ModulesConfig<Modules> | NoConfig,
+      RecursivePartial<ModulesConfig<Modules>>
+    >(this.currentConfig ?? {}, config);
   }
 
   // eslint-disable-next-line accessor-pairs
   public set config(config: ModulesConfig<Modules>) {
-    super.config = config;
-    this.definition.config = config;
+    console.log("Config before", this.currentConfig);
```
maht0rz: 👀

packages/common/src/config/ModuleContainer.ts
```diff
@@ -250,13 +264,24 @@ export class ModuleContainer<
...
+    super.config = merge<
+      ModulesConfig<Modules> | NoConfig,
+      ModulesConfig<Modules>
+    >(this.currentConfig ?? {}, config);
+    console.log("Config after", this.currentConfig);
```
maht0rz: 👀

PR #97, opened 2024-01-12
Block hooks pt. 2 aka new blockpipeline

rpanic:


.gitignore
```diff
@@ -39,4 +39,5 @@ testem.log
 Thumbs.db
 lerna-debug.log
 
-docs/save
\ No newline at end of file
+docs/save
+packages/research
```
maht0rz, line 43: i don't think we should creep personal habits into the repo, unless its something we use often and commonly

docs/sequencer/blockproduction.md
```diff
@@ -0,0 +1,55 @@
...
+1. StateTransition Reduction Pipeline:
+  This Step roughly takes an StateTransition[] and turns it into a proof attesting to the transformation from `stateRootA -> stateRootB` with a commitment to the StateTransitions `stateTransitionsHash`.
+  In the case there are more STs than the batch size, this is done recursively by merging the individual parts as a tree (a recurring pattern).
+2. Runtime Proof
+This is probably the easiers proof, it roughly proves a relation `Transaction -> ST[] via stateTransitionsHash`
+3. Transaction Proof
+This proof is actually part of the BlockProver circuit for technical reasons, but is essentially proves
+  `(stProof, runtimeProof) -> fromStateRoot, toStateRoot, tx[] via transactionsHash`. A Transaction[] is grouped into blocks, and for each block, it's transactions BlockProofs are merged into one
+4. BlockProof
+This proof proves takes a `txProof`, a `stProof` and executes a few steps
```
maht0rz, line 22: i think this isnt accurate since `proveBlock` takes in the _final merged block proof_ and runs the hooks at the end. i think even in the BlockProver.proveBlock the arg should be renamed from `transactionProof` to `blockProof` to be more accurate

packages/cli/src/commands/start.tsx
```diff
@@ -373,7 +373,7 @@ export function Server({ configFile }: { configFile: string }) {
           blockError,
           logs,
           height: block
-            ? Number(block?.networkState.block.height.toString())
+            ? Number(block?.networkState.during.block.height.toString())
```
maht0rz, line 376: i think we should call this `current` so it makes more sense within the runtime module API like `this.network.current.block.height`?

packages/common/src/trees/RollupMerkleTree.ts
```diff
@@ -205,6 +205,13 @@ export function createMerkleTree(height: number): AbstractMerkleTreeClass {
           ].toString()
         );
     }
+
+    public static dummy() {
```
maht0rz, line 209: -> `dummyWitness()`? or maybe implement this dummy() on the MerkleWitness itself

packages/protocol/src/protocol/ProvableBlockHook.ts
```diff
@@ -3,20 +3,17 @@ import { NetworkState } from "../model/network/NetworkState";
...
-export interface AfterBlockParameters {
-  state: BlockProverState;
-  networkState: NetworkState;
-}
-
-// Purpose is to validate transition from -> to network state
+// Purpose is to build transition from -> to network state
 export abstract class ProvableBlockHook<
   Config
 > extends TransitioningProtocolModule<Config> {
```
maht0rz, line 9: Why is this a `TransitioningProtocolModule` instead of e.g. ` BlockHookProtocolModule`?

packages/protocol/src/prover/block/BlockProvable.ts
```diff
@@ -5,19 +5,42 @@ import { StateTransitionProof } from "../statetransition/StateTransitionProvable
...
 
 export class BlockProverPublicOutput extends Struct({
   transactionsHash: Field,
   stateRoot: Field,
   networkStateHash: Field,
-}) {}
+  blockHashRoot: Field,
+  eternalTransactionsHash: Field,
+  closed: Bool,
+  blockNumber: Field,
```
maht0rz, line 26: Is this necessary to provide separately, since there is a network state commitment already? should be possible to witness network state and compare it with the commitment

packages/protocol/src/prover/block/BlockProver.ts
```diff
@@ -71,6 +77,14 @@ export interface BlockProverState {
    * This value is the same for the whole batch (L2 block)
    */
   networkStateHash: Field;
+
+  blockHashRoot: Field;
+
```
maht0rz, line 88: seems like unnecessary spaces

packages/protocol/src/prover/block/BlockProver.ts
```diff
@@ -234,81 +252,81 @@ export class BlockProverProgrammable extends ZkProgrammable<
...
+        } else if (type === "afterBlock") {
+          return blockHook.afterBlock(networkState, state);
+        } else {
+          throw new Error("Unreachable");
+        }
+      },
+      inputNetworkState
+    );
+
+    executionContext.afterMethod();
```
maht0rz, line 299: we should eventually add decorators to the block hook methods to avoid arbitrarily running the context methods like here

packages/protocol/src/prover/block/BlockProver.ts
```diff
@@ -319,51 +337,202 @@ export class BlockProverProgrammable extends ZkProgrammable<
...
-      bundleInclusionResult.networkState
+    transactionProof.publicInput.transactionsHash.assertEquals(
+      Field(0),
+      "TransactionProof transactionshash has to start at 0"
     );
-    const bundleClosed = executionData.transactionPosition.equals(
-      BlockTransactionPosition.fromPositionType("LAST")
+    transactionProof.publicInput.blockHashRoot.assertEquals(
+      Field(0),
+      "TransactionProof cannot carry the blockHashRoot - publicInput"
```
maht0rz, line 421: would be nice to move all errors to the errors definition at the top

packages/sdk/test/graphql/server.ts
```diff
@@ -105,7 +107,7 @@ export async function startServer() {
         BaseLayer: NoopBaseLayer,
         BlockProducerModule,
         UnprovenProducerModule,
-        BlockTrigger: ManualBlockTrigger,
+        BlockTrigger: TimedBlockTrigger,
```
maht0rz, line 110: why introduce the timed block trigger again? in my pov it makes it harder to test since you need to work with clocks, and the test becomes unnecessarily longer - unless you work with jest & clock spoofing

packages/sequencer/src/protocol/production/BlockProducerModule.ts
```diff
@@ -170,15 +173,14 @@ export class BlockProducerModule extends SequencerModule {
...
-      unprovenBlocks[0].block.networkState.block.height.toBigInt();
+    const blockId = unprovenBlocks[0].block.block.height.toBigInt();
 
     const block = await this.computeBlock(unprovenBlocks, Number(blockId));
 
     this.productionInProgress = false;
 
     const computedBundles = unprovenBlocks.map((bundle) =>
-      bundle.block.transactions.map((tx) => {
+      bundle.block.block.transactions.map((tx) => {
```
maht0rz, line 183: batch?

maht0rz (approved):
Please treat the comments as suggestions, which may be addressed at a later date

packages/cli/src/commands/start.tsx
rpanic: We don't have access to that in the runtime, there you only have `this.network`. The during is only something you can access in the query, and there I wanted to make sure to clearly highlight the network state that all the transactions used, and the two that are byproducts of the block pipeline

docs/sequencer/blockproduction.md
rpanic: Clarified it in the md why I think this makes sense and added a definition in there, as to not confuse readers and explain the background

packages/common/src/trees/RollupMerkleTree.ts
rpanic: It is implemented on the merklewitness itself, so I think this can be kept as other dummy() methods are also only called that without a suffix

packages/protocol/src/protocol/ProvableBlockHook.ts
rpanic: We already covered that some time ago on discord, but `TransitioningProtocolModule` allows ProtocolModule to use `State` and `StateMap` because it requires the implementation of a `name: string` property, which those classes need. And the ProvableBlockHook already is doing what you are describing in accordance with the namings we have for hooks

packages/protocol/src/prover/block/BlockProvable.ts
rpanic: We need this fields to make sure that during the merging of the `applyTransaction` proofs, we merge no proofs that are designated for different blocks and actually more importantly, to keep the integrity of the blockhashtree, where we need to make sure that the transaction-proofs cannot start a new leaf before the current one is closed, which we do via assertions on this value during merging. At least that was the original idea, but since the design has changed quite a bit since then, I will have to review this design choice again and see if we can remove it

packages/protocol/src/prover/block/BlockProver.ts
rpanic: 👍

packages/sdk/test/graphql/server.ts
rpanic: For this use case it makes sense, since this file is used by me to startup a production-like process where I can test the correct progression of things like networkstate and the behaviour of the mempool, etc. Which would be super tedious and actually impossible currently with the manual trigger

packages/sequencer/src/protocol/production/BlockProducerModule.ts
rpanic: No, its a block

PR #101, opened 2024-01-22
Feature/tx fees

maht0rz:


packages/library/src/hooks/TransactionFeeHook.ts
```diff
@@ -0,0 +1,255 @@
...
+      });
+    });
+
+    const witness = Provable.witness(FeeTree.WITNESS, () =>
+      this.feeTree.tree.getWitness(
+        this.feeTree.indexes[executionData.transaction.methodId.toString()]
+      )
+    );
+
+    const root = this.feeTree.tree.getRoot();
```
rpanic: I don't this it is a good idea to inline this logic that way for multiple reasons. It gets really hard to reason about what parts of that tree make it into the circuit and which don't. So I would like to see this logic (i.e. the tree + witness generation) in a decoupled unit. What I would do is to create a @singleton service consuming the runtime, building the tree and creating a getRoot() and getWitness() method. Also, I'd use bigint for getRoot() to make sure that there is no coupling and it is a constant

packages/library/src/hooks/TransactionFeeHook.ts
```diff
@@ -0,0 +1,255 @@
...
+      errors.invalidFeeConfigMethodId()
+    );
+
+    const fee = O1JSUInt64.from(feeConfig.baseFee.value).add(
+      O1JSUInt64.from(feeConfig.weight.value).mul(
+        O1JSUInt64.from(feeConfig.perWeightUnitFee.value)
+      )
+    );
+
+    const balances =
```
rpanic: Let's do that in the constructor. No particular reason but I get scared when I see circuit and non-circuit code mixed like this.

packages/library/src/runtime/Balances.ts
```diff
@@ -0,0 +1,126 @@
...
+      balanceOption.isSome,
+      Balance,
+      balanceOption.value,
+      Balance.from(0)
+    );
+  }
+
+  public setBalance(tokenId: TokenId, address: PublicKey, amount: Balance) {
+    const key = new BalancesKey({ tokenId, address });
+    this.events.emit("setBalance", key, amount);
```
rpanic: I am not really happy with that, since that requires a extra PR to work properly (would currently fire multiple times for different values without a way to really use them properly)

rpanic (approved):

packages/common/src/config/ModuleContainer.ts
```diff
@@ -346,9 +344,7 @@ export class ModuleContainer<
     moduleName: StringKeyOf<Modules>,
     containedModule: InstanceType<Modules[StringKeyOf<Modules>]>
   ) {
-    // Has to be super.config, getters behave really weird when subtyping
-    const config = super.config?.[moduleName];
-
+    const config = this.config?.[moduleName];
```
rpanic, line 351: This breaks the configuration functionality. Execute ContainerEvents.test.ts and see for yourself

PR #104, opened 2024-01-25
Fixed issues with UInts when using them with protokit state

rpanic:


packages/library/src/math/UInt.ts
```diff
@@ -195,12 +196,12 @@ export abstract class UIntX<This extends UIntX<any>> extends Struct({
     });
 
     rest
-      .rangeCheckHelper(this.NUM_BITS)
+      .rangeCheckHelper(this.numBits())
       .assertEquals(rest, "Sqrt rest output overflowing");
```
maht0rz, line 200: won't this bypass the assertion function?

maht0rz (approved):

packages/library/src/math/UInt.ts
rpanic: That's fine, since this assertion will only fail if the sequencer provides a malicious witness, in which case we actually want it to fail forcefully

PR #107, opened 2024-02-13
Fixes to the sdk tests with the new settlement & messaging

maht0rz:


rpanic (approved):
If the UInt stuff won't break anything, approved

PR #111, opened 2024-02-22
Implemented ST Merging

rpanic:
Spec in docs/st-merging.md


packages/protocol/src/settlement/SettlementContractModule.ts
```diff
@@ -0,0 +1,131 @@
...
+  public get appChain(): AreProofsEnabled | undefined {
+    return this.protocol?.getAreProofsEnabled();
+  }
+  // **
+
+  public create(childContainerProvider: ChildContainerProvider) {
+    super.create(childContainerProvider);
+  }
+
+  private assertThisFullfillsMinimumModules(): asserts this is SettlementContractModule<MinimumSettlementContracts> {
```
maht0rz: Can we use the mandatory modules approach we use for protocol also?

packages/protocol/src/model/StateTransitionReduction.ts
maht0rz, line 1: can this be co-located with the st hashlist implementation?

maht0rz (approved):

PR #122, opened 2024-04-08
updated to o1js 0.17.0

Raunaque97:
Update o1js version to 0.18.0 (latest)  [Part 1]

### DIFF
--- a/packages/common/test/zkProgrammable/ZkProgrammable.test.ts
+++ b/packages/common/test/zkProgrammable/ZkProgrammable.test.ts
@@ -61,7 +61,7 @@ class TestProgrammable extends ZkProgrammable<
 
   public zkProgramFactory() {
     const program = ZkProgram({
-      name: "",
+      name: "testZkProgram",
       publicInput: TestPublicInput,
       publicOutput: TestPublicOutput,
 
@@ -114,7 +114,7 @@ class OtherTestProgrammable extends ZkProgrammable {
 
   public zkProgramFactory(): PlainZkProgram {
     const program = ZkProgram({
-      name: "",
+      name: "plainZkProgram",
       methods: {
         bar: {
           privateInputs: [this.testProgrammable.zkProgram.Proof],
--- a/packages/common/test/zkProgrammable/ZkProgrammable.test.ts
+++ b/packages/common/test/zkProgrammable/ZkProgrammable.test.ts
@@ -91,9 +91,7 @@ class TestProgrammable extends ZkProgrammable<
       analyzeMethods: program.analyzeMethods.bind(program),
       Proof: SelfProof,
       methods,
-      analyzeMethods: () => {
-        return {};
-      },
+      analyzeMethods: program.analyzeMethods.bind(program),
     };
   }
 }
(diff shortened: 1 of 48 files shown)


packages/common/test/zkProgrammable/ZkProgrammable.test.ts
```diff
@@ -110,7 +114,8 @@ class OtherTestProgrammable extends ZkProgrammable {
   }
 
   public zkProgramFactory(): PlainZkProgram {
-    const program = Experimental.ZkProgram({
+    const program = ZkProgram({
+      name: "",
```
rpanic: Name is optional iirc, giving it "" might lead to collisions inside snarky

packages/common/test/zkProgrammable/ZkProgrammable.test.ts
```diff
@@ -89,6 +90,9 @@ class TestProgrammable extends ZkProgrammable<
       verify: program.verify.bind(program),
       Proof: SelfProof,
       methods,
+      analyzeMethods: () => {
```
rpanic: Please bind this like we did for verify() for example, we want to use that method eventually. Then we can also leave the type definition in ZkProgrammable as the correct type (ReturnType of the ZkProgram function). 
Also found in other parts of this PR

packages/protocol/src/utils/utils.ts
```diff
@@ -39,7 +39,7 @@ export function notInCircuit(): MethodDecorator {
 }
 
 export function stringToField(value: string) {
-  const fieldSize = Field.sizeInBytes() - 1;
+  const fieldSize = Field.sizeInBytes - 1;
```
rpanic, line 42: Did this API change or did we write that wrong in the first place?

packages/sdk/test/parameters.test.ts
```diff
@@ -22,7 +22,7 @@ import {
 import { TestingAppChain } from "../src/index";
 import { log } from "@proto-kit/common";
 import { assert, State, StateMap } from "@proto-kit/protocol";
-import { dummyBase64Proof } from "o1js/dist/node/lib/proof_system";
+import { dummyBase64Proof } from "o1js/dist/node/lib/proof-system";
```
rpanic: Afaik this should now be exposed by o1js, the dist import was only a workaround, which unfortunately broke the web deployments. But I think gregor exposed that since then, but correct me if I'm wrong

packages/sequencer/src/settlement/tasks/SettlementProvingTask.ts
```diff
@@ -8,6 +8,7 @@ import {
   SettlementContractModule,
 } from "@proto-kit/protocol";
 import { addCachedAccount, Field, Mina, Types } from "o1js";
+import { Transaction as MinaTransaction } from "o1js/dist/node/lib/mina";
```
rpanic: dist import :(
I think the Transaction objects are now Transaction, PendingTransaction and IncludedTransaction. See what fit here best

rpanic: Ah yeah you got the Transaction right already, then only the import has to be changed

packages/sdk/test/parameters.test.ts
rpanic: Its Proof.dummy()

Raunaque97: I updated it, Thanks for the info

packages/protocol/src/utils/utils.ts
Raunaque97: I think it changed with the new version

PR #175, opened 2024-07-18
Unit-tests for Transaction Fees

### LINKED ISSUE #143: Improve test coverage for transaction fees
We need better test coverage for transaction fees - we dont have a metric for test coverage as of now, so adding additional test cases and "trying to break" the transaction fee implementation should be the focus.

N/A

**Additional context**
N/A

ejMina226:
New tests added:

1. Transaction is rejected when sender can't cover the fee
2. Check fee-payer receives fee when it is subtracted.
3. Check app chain doesn't start when config is bad for transaction-fee.
4. Add check that fee was computed properly, including helper function in TransactionFeeHook.ts

Also,
1.  Fixed typo: changing `intializeFeeTree` to  `initializeFeeTree`
2. Added in nvmrc file


packages/library/src/hooks/TransactionFeeHook.ts
```diff
@@ -111,6 +111,13 @@ export class TransactionFeeHook extends ProvableTransactionHook<TransactionFeeHo
     );
   }
 
+  public getFee(methodId: bigint) {
```
rpanic: Ideally I'd like this method being used by the onTransaction hook itself. Currently it implements the same thing.
The only difference there is that it uses UInt64, which this method probably should too, otherwise you could get some discrepancies between tests and the runtime

packages/sdk/test/fees-failures.test.ts
```diff
@@ -0,0 +1,171 @@
...
+    await this.balances.mint(
+      TokenId.from(0),
+      this.transaction.sender.value,
+      Balance.from(1000)
+    );
+  }
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+interface RuntimeModules extends RuntimeModulesRecord {
```
rpanic: For correctness, this should inherit from `VanillaRuntimeModulesRecord` as that includes the Balances as well

packages/sdk/test/fees.test.ts
```diff
@@ -115,27 +116,73 @@ describe("fees", () => {
...
+    const burnAmount = Balance.from(100);
+    const senderBalanceBefore = balanceSenderBefore;
 
     const tx = await appChain.transaction(senderKey.toPublicKey(), async () => {
-      await pit.burn(Balance.from(100));
+      await pit.burn(burnAmount);
     });
 
+    const methodId = tx.transaction?.methodId.toBigInt();
+    if (methodId === undefined) {
```
rpanic: Any reason you did that this way? Otherwise this should be expectDefined as well

packages/sdk/test/fees-failures.test.ts
ejMina226: Thanks. I just lifted this from the fees.test.ts set up.

ejMina226: What is the point of this interface? it doesn't look to be used anywhere?

packages/sdk/test/fees.test.ts
ejMina226: No reason. Thanks.

packages/library/src/hooks/TransactionFeeHook.ts
ejMina226: Are you saying this function should be moved into the onTransaction function in the same file. If so, how can I call it in the test? Thanks.

packages/sdk/test/fees.test.ts
```diff
@@ -115,27 +116,71 @@ describe("fees", () => {
...
+    const senderBalanceBefore = balanceSenderBefore;
 
     const tx = await appChain.transaction(senderKey.toPublicKey(), async () => {
-      await pit.burn(Balance.from(100));
+      await pit.burn(burnAmount);
     });
 
+    const methodId = tx.transaction?.methodId.toBigInt();
+    expectDefined(methodId);
+    const transactionFee = transactionFeeModule.getFee(methodId);
```
ejMina226: @rpanic  Do you know why intellisense would be throwing an error on this line? The code runs fine, it's just the intellisense. The message says: `Unsafe assignment of an error typed value.eslint[@typescript-eslint/no-unsafe-assignment](https://typescript-eslint.io/rules/no-unsafe-assignment)`

packages/library/src/hooks/TransactionFeeHook.ts
rpanic: No i meant that the onTransaction logic should call this function internally. Since it has the same logic inside to calculate the fee, therefore it should be unified by both the test and onTransaction calling getFee()

packages/sdk/test/fees-failures.test.ts
rpanic: Which interface concretely? We have interfaces like these all over the place, as they indicate what shape the modules record for a given container has. The RuntimeModulesRecord is only a generic mapping of string => RuntimeModule, which will be extended by the Runtime or by extending it manually for reusable module bundles.
These Records generally are used for the resolving modules (via runtime.resolve("abc") for example) or for infering the configs

ejMina226: So the RuntimeModules interface. It's not used anywhere explicitly in this file. Also, when I used the RuntimeModulesRecord the code didn't error, so why is the VanillaRuntimeModulesRecord necessary? Thanks.

rpanic: Ah yes, if its not used, feel free to delete it

packages/library/src/hooks/TransactionFeeHook.ts
ejMina226: Sure. Makes sense. Is this what you were after? I've pushed an update.

rpanic: Yes, but you need to make sure it's provable end to end. That also goes for the methodId, you can't use bigints in provable code. Other that that, it's what I meant, yeah

ejMina226: Ah, OK. So are you saying I can't use the methodId, or that I have to use the Provable.witness inside the function to mitigate the methodId? Thanks.

rpanic (approved):

PR #177, opened 2024-07-19
Fixed error for 0-length runtime args

### LINKED ISSUE #70: Using 0 arguments in an init runtimeMethod under a runtimeModule returns an error.
**Describe the bug**
When sending the first txn to init a runtimeModule, if we don't have any function parameter/argument, it returns an error as :
```
    Error in worker: argsHash and therefore arguments of transaction and runtime call does not match
    Field.assertEquals(): 21565680844461314807147611702860246336805372493508489110556896454939225549736 != 0
```

**To Reproduce**
1. Follow the steps from the starter-kit.

2. Update the Balances.ts as :
```
import { RuntimeModule, runtimeModule, runtimeMethod } from "@proto-kit/module";

interface BalancesConfig {}

@runtimeModule()
export class Balances extends RuntimeModule<BalancesConfig> {
  @runtimeMethod()
  public init() {}
}
```

3. Update the Balances.test.ts to:
```
import { TestingAppChain } from "@proto-kit/sdk";
import { PrivateKey } from "o1js";
import { Balances } from "./Balances";

describe("Balances", () => {
  it("should demonstrate how balances work", async () => {
    const appChain = TestingAppChain.fromRuntime({
      modules: {
        Balances,
      },
      config: {
        Balances: {},
      },
    });

    await appChain.start();

    const alicePrivateKey = PrivateKey.random();
    const alice = alicePrivateKey.toPublicKey();

    appChain.setSigner(alicePrivateKey);

    const balances = appChain.runtime.resolve("Balances");

    const tx1 = appChain.transaction(alice, () => {
      balances.init();
    });

    await tx1.sign();
    await tx1.send();

    await appChain.produceBlock();
  });
});
```

4. In the terminal run `npm run test:watch`

**Expected behavior**
Should init the runtimeModule and the tests should pass.

**Desktop (please complete the following information):**
 - OS: Ubuntu 22
 - Protokit: Latest - 28/10/23

rpanic:
Closes #70  

Merge #174 first

This PR fixes a bug that occurred when a runtime method had no parameters. 
This was due to a discrepancy in the argshash construction for the transaction objects (and therefore signature) and the in-circuit runtime method re-hashing of the args. The runtime method one defaulted to Field(0) for empty args, whereas the transaction one didn't. Unified to doing `hash([])` for empty args

### DIFF
--- a/packages/module/src/method/runtimeMethod.ts
+++ b/packages/module/src/method/runtimeMethod.ts
@@ -112,11 +112,7 @@ export function toWrappedMethod(
     ).encode(args);
 
     // Assert that the argsHash that has been signed matches the given arguments
-    // We can use js-if here, because args are statically sized
-    // i.e. the result of the if-statement will be the same for all executions
-    // of this method
-    const argsHash =
-      (args ?? []).length > 0 ? Poseidon.hash(argsFields) : Field(0);
+    const argsHash = Poseidon.hash(argsFields);
 
     transaction.argsHash.assertEquals(
       argsHash,
--- a/packages/sequencer/test/integration/BlockProduction.test.ts
+++ b/packages/sequencer/test/integration/BlockProduction.test.ts
@@ -120,41 +120,6 @@ describe("block production", () => {
     mempool = sequencer.resolve("Mempool");
   });
 
-  it.only("regression - should produce block with no STs emitted", async () => {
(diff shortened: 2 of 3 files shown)


ejMina226 (approved):
LGTM. What is noops runtime doing?

PR #178, opened 2024-07-23
Unit test to confirm transfer from and to same account does not error

### LINKED ISSUE #173: Fix Balances.transfer issue with tokens being minted when from === to

ejMina226:


packages/sdk/test/minting-error.test.ts
```diff
@@ -0,0 +1,121 @@
...
+    await this.balances.mint(
+      TokenId.from(0),
+      this.transaction.sender.value,
+      Balance.from(1000)
+    );
+  }
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+interface RuntimeModules extends RuntimeModulesRecord {
```
rpanic: Same thing, should be `VanillaRuntimeModulesRecord`

rpanic (approved):
lgtm!

packages/sdk/test/minting-error.test.ts
ejMina226: It's not used so I'll remove it.

PR #181, opened 2024-07-24
Unify transaction argument encoding

### LINKED ISSUE #154: Unify transaction argument encoding
Currently we support two encoding mechanisms: JSON & field[]. This is due to historic reasons where we were not able to effectively implement the field encoding due to internal o1js discrepancies.

The JSON encoding should be removed in favour of the field encoding, but at the same time the field encoding should be thoroughly tested on various o1js data types. 

Proofs which can be used as arguments, cannot be encoded as fields. However proof inputs/outputs can be encoded as fields since they are provable, and part of the encoding mechanism & the `transactionHash` commitment used by the Runtime circuits. This means that we have to transport the raw proofs in some additional attributes of the transaction model to maintain the same functionality as with the previous JSON encoding.

First pass at testing the encoding is done here: https://github.com/proto-kit/framework/blob/develop/packages/sdk/test/parameters.test.ts

N/A

**Additional context**
N/A

rpanic:
Closes #154 

Based on #177 (unfortunately)

This PR removes the json-encoding of transaction arguments and replaces it by a tandem of encoding args as Fields and auxiliary data.
Arguments that are provable on their own, plus public inputs and outputs of Proofs will be encoded as a `Field[]`.
The proofs themselves will be encoded as `auxilaryData` which is a `string[]` (since not provable)


ejMina226 (approved):
LGTM

PR #185, opened 2024-07-26
Provable Events

### LINKED ISSUE #114: RFC: Runtime events
A often requested feature is to be able to emit provable events inside runtime methods.
Inspiration is probably coming from solidity and zkapps, where events already exist with similar use-cases.

Generally speaking events are metadata about a certain execution of a transaction that can be read afterwards. These events are dependent (and in our case proven) to be emitted by some values that are computed during a method's execution and cannot be tampered with. 

From my perspective, the high level usages include:
- Direct usage in UIs
- Tracking of transaction results without re-executing a given transaction
- More friendly and efficient metadata extraction for indexer & processor
- Easier application testing without having to fetch the raw state from the storage

### Suggested API

I kept closely to the zkapps API (which is also very similar to the solidity API) for this design
A sample API could look like

```ts
export class TransferEvent extends Struct({
  from: PublicKey,
  to: PublicKey,
  amount: UInt64,
}){}

@runtimeModule()
export class Balances extends RuntimeModule {
  @state() public balances = StateMap.from<PublicKey, UInt64>(
    PublicKey,
    UInt64
  );

  public events = new RuntimeEvents({
    transfer: TransferEvent
  })
  
  @runtimeMethod
  public transfer(from: PublicKey, to: PublicKey, amount: UInt64) {
    ...
    this.events.emit("transfer", new TransferEvent({ from, to, amount }))
  }
}
```

As we can see, first we define the type of the Event as a struct.
Then, we define the event name and it's type on the RuntimeModule, which we can then later use to emit a event inside of a runtime method.

### Implementation

The way this manifests in the proving cycle and the proof outputs is the following.
The emitted events will be captured by the ExecutionContext, which will then be used to retrieve the raw event data (like we do with STs). Additionally, a commitment to the emitted events is attached to the RuntimeProof.publicOutput.

This commitment is computed as a hashlist of the hash of each event.
I.e. something like `hash(events: Provable[]) => Poseidon.hash(...events[0].toFields(), hash(events.slice(1)))`.
This could also be simplified by just hashing them together as a single hash (since it is statically sized in principle).

Apart from the Runtimeproof having a commitment to the events attached, further processing this commitment isn't necessary, because the event output isn't usable anywhere as new inputs, and they have no actions attached to them (for that, we built messages).

Why do we define the events object on the class and why do't we allow arbitrary event emittance by using .toFields() on the event objects?
Because that way we can type-check the emitting of events and also the querying of events. Also, a fixed set of events with their respective type layout seems to be important for the indexer and processor to know what can be attached to a transaction. 

### Open Questions

Should we add a emit variant that allows to conditionally send an event? Something like
`this.emitIf(enoughBalance, "transfer", transferEvent)`

ejMina226:
Add in provable events into the codebase. This required changes to 
- the RuntimeModule class
- the RuntimeMethodExecution class
- the TransactionMapper (for the persistent storage)
- testing (i.e. unit test, `BlockProduction.test`, `PrismaBlockProduction.test`)

Closes #114

### DIFF
--- a/packages/module/src/method/runtimeMethod.ts
+++ b/packages/module/src/method/runtimeMethod.ts
@@ -63,10 +63,7 @@ export function toEventsHash(
     .map((event) => {
       return event.eventType.toFields(event.event);
     })
-    .reduce(
-      (accum, next) => Poseidon.hash([accum, ...next]),
-      Poseidon.hash([Field(0)])
-    );
+    .reduce((accum, next) => Poseidon.hash([accum, ...next]), Field(0));
 }
 
 export type WrappedMethod = (...args: ArgumentTypes) => MethodPublicOutput;
--- a/packages/module/src/method/runtimeMethod.ts
+++ b/packages/module/src/method/runtimeMethod.ts
@@ -122,13 +122,13 @@ export function toWrappedMethod(
      * Use the type info obtained previously to convert
      * the args passed to fields
      */
-    const { argsFields } = MethodParameterEncoder.fromMethod(
+    const { fields } = MethodParameterEncoder.fromMethod(
       this,
       methodName
     ).encode(args);
 
     // Assert that the argsHash that has been signed matches the given arguments
-    const argsHash = Poseidon.hash(argsFields);
+    const argsHash = Poseidon.hash(fields);
 
     transaction.argsHash.assertEquals(
       argsHash,
--- a/packages/module/src/method/runtimeMethod.ts
+++ b/packages/module/src/method/runtimeMethod.ts
@@ -1,4 +1,4 @@
-import { Bool, Field, Poseidon } from "o1js";
+import { Bool, Field, FlexibleProvablePure, Poseidon } from "o1js";
 import { container } from "tsyringe";
 import {
   StateTransition,
@@ -52,6 +52,10 @@ export function toStateTransitionsHash(
     .toField();
 }
 
+export function toEventsHash(events: FlexibleProvablePure<any>[]) {
+  return events.map((event) => Poseidon.hash(Field([event]))).toField();
+}
+
 export type WrappedMethod = (...args: ArgumentTypes) => MethodPublicOutput;
 export type AsyncWrappedMethod = (
   ...args: ArgumentTypes
@@ -74,10 +78,11 @@ export function toWrappedMethod(
   ): Promise<MethodPublicOutput> => {
     await Reflect.apply(moduleMethod, this, args);
     const {
-      result: { stateTransitions, status },
+      result: { stateTransitions, status, events },
     } = executionContext.current();
 
     const stateTransitionsHash = toStateTransitionsHash(stateTransitions);
+    const eventsHash = toEventsHash(events);
 
     const { name, runtime } = this;
 
@@ -131,6 +136,7 @@ export function toWrappedMethod(
       transactionHash,
       networkStateHash,
       isMessage,
+      eventsHash,
     });
   };
 
--- a/packages/module/src/method/runtimeMethod.ts
+++ b/packages/module/src/method/runtimeMethod.ts
@@ -53,7 +53,7 @@ export function toStateTransitionsHash(
 }
 
 export function toEventsHash(events: FlexibleProvablePure<any>[]) {
-  return events.map((event) => Poseidon.hash(Field([event]))).toField();
+  return Poseidon.hash(events.map((event) => Field([event])));
 }
 
 export type WrappedMethod = (...args: ArgumentTypes) => MethodPublicOutput;
--- a/packages/module/src/method/runtimeMethod.ts
+++ b/packages/module/src/method/runtimeMethod.ts
@@ -61,6 +61,9 @@ export function toEventsHash(
     condition: Bool;
   }[]
 ) {
+  if (events.length === 0) {
+    return Field(0);
+  }
   const hashList = new DefaultProvableHashList(events[0].eventType);
   events.forEach((event) => {
     hashList.pushIf(event.event, event.condition);
--- a/packages/module/src/method/runtimeMethod.ts
+++ b/packages/module/src/method/runtimeMethod.ts
@@ -61,6 +61,7 @@ export function toEventsHash(
 ) {
   return events
     .map((event) => {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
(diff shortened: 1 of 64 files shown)


packages/module/src/method/runtimeMethod.ts
```diff
@@ -52,6 +52,21 @@ export function toStateTransitionsHash(
...
 }
 
+export function toEventsHash(
+  events: {
+    eventType: FlexibleProvablePure<any>;
+    event: any;
+    eventName: string;
+  }[]
+) {
+  return events
```
rpanic: We already have a abstraction for this sort of behaviour, that I kept from you. It's called `DefaultProvableHashList`

packages/persistance/src/services/prisma/mappers/EventMapper.ts
```diff
@@ -0,0 +1,63 @@
...
+export class EventMapper
+  implements
+    ObjectMapper<{ eventName: string; data: Field[] }, Prisma.JsonObject>
+{
+  public mapIn(input: Prisma.JsonObject): { eventName: string; data: Field[] } {
+    if (input === undefined) return { eventName: "", data: [] };
+    return {
+      eventName: input.eventName as string,
+      data: (input.data as Prisma.JsonArray).map((field) =>
+        Field.fromJSON(field as string)
```
rpanic, line 17: Field(field)

packages/persistance/test-integration/utils.ts
```diff
@@ -65,11 +65,20 @@ export const IntegrationTestDBConfig = {
   redisConfig,
 };
 
+export class TestEvent extends Struct({
+  message: Bool,
+}) {}
+
 @runtimeModule()
 export class MintableBalances extends Balances {
+  public events = {
```
rpanic: I forgot to also change that along with my changes, need to be changed as well pls :)

packages/sequencer/test/integration/BlockProduction.test.ts
```diff
@@ -30,10 +35,31 @@ import { ProtocolStateTestHook } from "./mocks/ProtocolStateTestHook";
...
+  message: Bool,
+}) {}
+
+@runtimeModule()
+class EventMaker extends RuntimeModule<unknown> {
+  public constructor() {
+    super();
+  }
+
+  public events = {
```
rpanic: Same here :/

rpanic (changes requested):
good job!

packages/persistance/src/services/prisma/mappers/EventMapper.ts
ejMina226: @rpanic Typescript is complaining with this change, saying 
`Argument of type 'JsonValue' is not assignable to parameter of type 'string | number | bigint | Field | FieldConst | FieldVar'.
  Type 'null' is not assignable to type 'string | number | bigint | Field | FieldConst | FieldVar'.t`

packages/module/src/method/runtimeMethod.ts
ejMina226: It's not clear to me that this is relevant. We could have different event types in the list of events and therefore we will need to create different instances of each DefaultProvableHashList, right? Or have I misunderstood? Thanks.

ejMina226: If we can just reuse the same ProvableType then we can do: 
`  const hashList = new DefaultProvableHashList(events[0].eventType);
  events.forEach((event) => {
    hashList.push(event.event);
  });
  return hashList.commitment; `

packages/persistance/test-integration/utils.ts
ejMina226: I've updated the test to reflect the new format. However, there appears to be a linting error here
`  public events = new RuntimeEvents({
    test: TestEvent,
  });`

The error is: `Unsafe assignment of an error typed value.`

packages/module/src/runtime/RuntimeModule.ts
```diff
@@ -17,6 +17,57 @@ const errors = {
...
+      throw new Error("Only string");
+    }
+    return container
+      .resolve(RuntimeMethodExecutionContext)
+      .addEvent(eventType, event, eventName);
+  }
+
+  // eslint-disable-next-line consistent-return
+  public emitIf<Key extends keyof Events>(
+    condition: emitIfConditional,
```
ejMina226: @rpanic I've added in an implementation of the emitIf.

packages/module/src/method/runtimeMethod.ts
ejMina226: Resolved in Slack thread.

packages/persistance/src/services/prisma/mappers/EventMapper.ts
rpanic: Okay then let's leave it, they should be semantically the same anyways

packages/module/test/runtimeMethod.test.ts
```diff
@@ -76,4 +133,82 @@ describe("runtimeMethod", () => {
...
+    expect(eventsResults[2]).toStrictEqual(thirdExpectedEvent);
+
+    context.afterMethod();
+
+    const proof = await context.current().result.prover!();
+    const publicOuput = proof.publicOutput as MethodPublicOutput;
+    const { eventsHash } = publicOuput;
+    //  Note that we omit the second event from below as it was
+    //  not emitted due to the condition being false.
+    expect(eventsHash).toStrictEqual(
```
rpanic, line 207: Thing here is that toEventsHash has to respect the condition on it's own, because in-circuit we will call toEventsHash with all events every time. We implement that correctly, but the test should also reflect that, i.e. we should feed in all 3 events here

packages/module/src/runtime/RuntimeModule.ts
```diff
@@ -17,6 +17,41 @@ const errors = {
...
+type EventRecord = Record<string, FlexibleProvablePure<any>>;
+
+type InferProvable<T extends FlexibleProvablePure<any>> =
+  T extends Provable<infer U> ? U : never;
+
+export class RuntimeEvents<Events extends EventRecord> {
+  public constructor(private readonly events: Events) {}
+
+  public emitIf<Key extends keyof Events>(
+    eventName: Key,
```
rpanic, line 30: As discussed, lets make the condition the first argument

rpanic (approved):

PR #195, opened 2024-09-02
Use Sideloading for runtime methods

### LINKED ISSUE #190: Use Sideloading for runtime methods
Currently, wrap circuits in pickles have a maximum number of circuit methods that can be used. This limit is quite low and users are expected to run into this limitation eventually. Additionally, compilation of any circuit currently requires all circuits it consumes to be compiled as well which leads to bad memory performance. 

o1js now exposes sideloading, which can be used to split up zkprograms into multiple pieces and then stitch them together afterwards.
The propsed solution is the following:
Create a sepearte ZkProgram for each runtime method. In parallel, hash the verification keys of all methods and put them into a merkle tree, whose root will then be baked into the BlockProver circuit via a Field constant. 
For workers, proving becomes really easy, compile (or load) the circuit for the specific method they have to prove, and return the proof to the sequencer. 
During `proveTransaction`, the BlockProver will then prove the membership of this proof's verification key in the global runtime-method merkle tree by providing a tree witness to that and computing the root using the verification key's hash (which is a builtin primitive in sideloaded keys). The resulting root can then be asserted to equal the previously baked in root in the BlockProver. 
This procedure ensures that only valid runtime methods can be proven since otherwise the verification key would change (since the methodId is also baked into the runtime method's circuit)

ejMina226:
See https://github.com/proto-kit/framework/issues/190


packages/protocol/src/prover/block/BlockProver.ts
```diff
@@ -154,13 +174,15 @@ export class BlockProverProgrammable extends ZkProgrammable<
...
-    executionData: BlockProverExecutionData
+    executionData: BlockProverExecutionData,
+    verificationKey: VerificationKey
   ): Promise<BlockProverState> {
     const { transaction, networkState, signature } = executionData;
 
     const { isMessage } = runtimeProof.publicOutput;
 
-    runtimeProof.verify();
+    const runtimeProofDynamic = DynamicRuntimeProof.fromProof(runtimeProof);
```
rpanic: This usage is incorrect. What I meant by transforming a Proof into a DynamicProof was that that needs to happen outside of the circuit before the execution. Because the type of Proof class is what o1js uses to distinguish baked-in proofs and sideloaded ones. Therefore, this service should entirely deal with Dynamic Proofs for the Runtime Proofs

packages/protocol/src/prover/block/BlockProver.ts
```diff
@@ -746,11 +800,10 @@ export class BlockProverProgrammable extends ZkProgrammable<
   public zkProgramFactory(): PlainZkProgram<
     BlockProverPublicInput,
     BlockProverPublicOutput
-  > {
+  >[] {
     const { prover, stateTransitionProver, runtime } = this;
-    const StateTransitionProofClass = stateTransitionProver.zkProgram.Proof;
-    const RuntimeProofClass = runtime.zkProgram.Proof;
-
+    const StateTransitionProofClass = stateTransitionProver.zkProgram[0].Proof;
```
rpanic, line 784: Yeah this doesn't make much sense does it. Keep the above comment in mind here, we need to enable dynamically linking of the different shards of the runtime here. We do this by using DynamicProofs here, that can take any shape of any on the shards. Later in the program, we assert their validity with the already existing logic

packages/sequencer/src/protocol/production/tasks/BlockProvingTask.ts
```diff
@@ -124,7 +124,7 @@ export class BlockProvingTask
   private readonly blockProver: BlockProvable;
 
   private readonly runtimeProofType =
```
rpanic, line 135: Again, this here needs to become a DynamicProof and shouldn't have any reference to the runtime, only the reference to the tree (and service) should be needed

packages/protocol/src/prover/block/BlockProver.ts
ejMina226: I'm taking the first element because I was told that the general shape should be the same for each zkProgramProof. Are you saying this isn't the case and I need to use the VerificationKeyService to extract the right proof? Thanks.

ejMina226: Resolved in Slack thread.

packages/library/src/hooks/RuntimeFeeAnalyzerService.ts
```diff
@@ -80,55 +80,67 @@ export class RuntimeFeeAnalyzerService extends ConfigurableModule<RuntimeFeeAnal
...
-            ...indexes,
-            [methodId.toString()]: BigInt(index),
-          },
-        ];
-      },
-      [{}, {}]
-    );
+            [{}, {}]
+          );
+          const [valuesProg, indexesProg] = await accum;
```
rpanic: I think awaiting the accum should be in the first line of the reduce, so that things don't get in the way of each other and we can guarantee sequential execution

packages/protocol/src/prover/block/BlockProver.ts
```diff
@@ -116,6 +129,12 @@ function maxField() {
 export type BlockProof = Proof<BlockProverPublicInput, BlockProverPublicOutput>;
 export type RuntimeProof = Proof<void, MethodPublicOutput>;
 
+class DynamicRuntimeProof extends DynamicProof<Void, MethodPublicOutput> {
```
rpanic: Lets reuse the one that is defined in BlockProvable here

packages/protocol/src/prover/block/BlockProver.ts
```diff
@@ -397,6 +418,36 @@ export class BlockProverProgrammable extends ZkProgrammable<
...
+    const calculatedRoot = witness.calculateRoot(zkProgramConfig.hash());
+    root.assertEquals(calculatedRoot, errors.invalidZkProgramTreeRoot());
+
+    zkProgramConfig.methodId.assertEquals(
+      executionData.transaction.methodId,
+      errors.invalidZkProgramConfigMethodId()
+    );
+    vkRecord.hash.assertEquals(zkProgramConfig.vkHash);
+
+    runtimeProof.verify(vkRecord);
```
rpanic: We verify the runtimeproof twice here - but I already fixed it in my branch just for you to be aware of it

packages/sequencer/src/protocol/production/tasks/BlockProvingTask.ts
```diff
@@ -65,6 +65,12 @@ export class DecodedStateSerializer {
   }
 }
 
+class DynamicRuntimeProof extends DynamicProof<Void, MethodPublicOutput> {
```
rpanic: Again the same proof definition

rpanic (approved):

PR #196, opened 2024-09-09
Protokit indexer

maht0rz:
This PR introduces the app-chain indexer, acting as an archive node for the sequencer. I did not implement pruning of historical blocks on the sequencer, we need to do that separately. Rationale here is that we want to keep the sequencer DB lean and small, so it can operate fast.

On the other hand the indexer can afford to store a lot of historical data - as in all historical blocks and transactions.

The indexer can be found in `packages/indexer`, and contains the following components:
- `GeneratedResolverFactoryGraphqlModule`
  - Auto-generated GQL resolvers based on the Indexer's prisma schema, using `typegraphql-prisma`
- `IndexBlockTask.ts`
  - Task for processing the inbound blocks from the sequencer and storing them in the indexer's storage services
- `IndexerNotifier.ts`
  - Sequencer module for listening to block production events, and pushing indexing tasks to the task queue for the indexer to pick up

Additional changes done to the `packages/api`:
- `ResolverFactoryGraphqlModule`
  - Introduced a new graphql module type, responsible for returning an array of predefined `resolvers()`. This is helpful for returning an array of automatically generated resolvers


ejMina226 (approved):
LGTM.

PR #205, opened 2024-10-03
Improve Transaction Validation in the Mempool

### LINKED ISSUE #163: Improve transaction validation in the mempool

**Additional context**

- Verify proofs as argument, throw away the tx any proof doesn't verify
- For non-proven appchain deployments, we still need to compile the requires proof, in case it's not a DynamicProof, to be able to validate proofs-as-arguments against that resulting VK. This means that users will have to still produce the proofs that they send to the runtime, but otherwise every use can send unrestricted public input/outputs to the runtime and potentially break the system

ejMina226:
Improvements include:
1. Getting txs from the mempool no longer ignores earlier txs that are validated by later ones (e.g. nonce misalignment)


packages/sequencer/src/mempool/private/PrivateMempool.ts
```diff
@@ -9,17 +18,38 @@ import {
...
 import { TransactionValidator } from "../verification/TransactionValidator";
+import { BlockQueue } from "../../storage/repositories/BlockStorage";
+import {
+  Sequencer,
+  SequencerModulesRecord,
+} from "../../sequencer/executor/Sequencer";
+import { CachedStateService } from "../../state/state/CachedStateService";
+import { AsyncStateService } from "../../state/async/AsyncStateService";
 
+type MempoolStateTransition = {
```
rpanic: I this that name isn't fitting, something like MempoolTransactionPaths makes more sense

packages/sequencer/src/mempool/private/PrivateMempool.ts
```diff
@@ -50,8 +80,98 @@ export class PrivateMempool extends SequencerModule implements Mempool {
     );
   }
 
+  private get unprovenQueue(): BlockQueue {
```
rpanic: Here we should not use the queue but the `BlockStorage` interface, it exposes a method `getLatestBlock()`

packages/sequencer/src/mempool/private/PrivateMempool.ts
```diff
@@ -50,8 +80,98 @@ export class PrivateMempool extends SequencerModule implements Mempool {
...
+  }
+
+  public async getStagedNetworkState(): Promise<NetworkState | undefined> {
+    const result = await this.unprovenQueue.getLatestBlock();
+    return result?.result.afterNetworkState;
+  }
+
   public async getTxs(): Promise<PendingTransaction[]> {
-    return await this.transactionStorage.getPendingUserTransactions();
+    const txs = await this.transactionStorage.getPendingUserTransactions();
```
rpanic, line 97: Can you decrease the cognitive complexity of this function and refactor it a bit so that its not 80 lines long?

packages/sequencer/src/mempool/private/PrivateMempool.ts
```diff
@@ -50,8 +80,98 @@ export class PrivateMempool extends SequencerModule implements Mempool {
...
+    const sortedTxs: PendingTransaction[] = [];
+    const skippedTxs: Record<string, MempoolStateTransition> = {};
+    const executionContext = container.resolve<RuntimeMethodExecutionContext>(
+      RuntimeMethodExecutionContext
+    );
+    const baseCachedStateService = new CachedStateService(this.stateService);
+    this.protocol.stateServiceProvider.setCurrentStateService(
```
rpanic: You can skip that step, since we don't run any state-accessing code in between here and the next setCurrentS() call. Therefore setting it here isn't necessary

packages/sequencer/src/mempool/private/PrivateMempool.ts
```diff
@@ -50,8 +80,98 @@ export class PrivateMempool extends SequencerModule implements Mempool {
...
+      baseCachedStateService
+    );
+    const networkState =
+      // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+      (await this.getStagedNetworkState()) as NetworkState;
```
rpanic: Add a proper undefined check here instead of casting

packages/sequencer/src/mempool/private/PrivateMempool.ts
rpanic: Also, can you add a few comments here and there to document the algorithms and process you use here?

packages/sequencer/src/mempool/private/PrivateMempool.ts
```diff
@@ -50,8 +80,98 @@ export class PrivateMempool extends SequencerModule implements Mempool {
...
+        // eslint-disable-next-line no-await-in-loop
+        await this.accountStateHook.onTransaction({
+          networkState: networkState,
+          transaction: signedTransaction.transaction,
+          signature: signedTransaction.signature,
+        });
+        const { status, statusMessage, stateTransitions } =
+          executionContext.current().result;
+        if (status.toBoolean()) {
+          console.log(`Accepted tx ${tx.hash().toString()}`);
```
rpanic: console.log -> log.*

packages/sequencer/src/mempool/private/PrivateMempool.ts
```diff
@@ -50,8 +80,98 @@ export class PrivateMempool extends SequencerModule implements Mempool {
...
+          }
+        } else {
+          console.log(
+            `Skipped tx ${tx.hash().toString()} because ${statusMessage}`
+          );
+          this.protocol.stateServiceProvider.popCurrentStateService();
+          if (!reEvaluation) {
+            skippedTxs[tx.hash().toString()] = {
+              transaction: tx,
+              paths: stateTransitions.map((x) => x.path),
```
rpanic: Add a distinct filter here so we don't save duplicate paths

packages/sequencer/src/mempool/private/PrivateMempool.ts
```diff
@@ -50,8 +80,98 @@ export class PrivateMempool extends SequencerModule implements Mempool {
...
+          // eslint-disable-next-line no-await-in-loop
+          await txStateService.applyStateTransitions(stateTransitions);
+          // eslint-disable-next-line no-await-in-loop
+          await txStateService.mergeIntoParent();
+          this.protocol.stateServiceProvider.popCurrentStateService();
+          delete skippedTxs[tx.hash().toString()];
+          if (Object.entries(skippedTxs).length > 0) {
+            const candidateMissedTxs: PendingTransaction[] = [];
+            // eslint-disable-next-line array-callback-return
+            stateTransitions.map((st) => {
```
rpanic: foreach, eslint also nudges you to use it I assume

rpanic: You can alternatively use map and return the tx only when it has a matching path, then do `filter(filterNonUndefined)`

packages/sequencer/src/mempool/private/PrivateMempool.ts
```diff
@@ -50,8 +80,98 @@ export class PrivateMempool extends SequencerModule implements Mempool {
...
+              Object.values(skippedTxs).forEach((value) => {
+                if (value.paths.some((x) => x.equals(st.path))) {
+                  candidateMissedTxs.push(value.transaction);
+                }
+              });
+            });
+            if (candidateMissedTxs.length > 0) {
+              // eslint-disable-next-line no-await-in-loop
+              await checkTxValid(candidateMissedTxs, true);
```
rpanic: I think this is also not the best thing to do since if we have multiple new candidates with the same path, this will execute it multiple times.
Imagine txs from a common sender in the order `[{ nonce: 2}, {nonce:3}, {nonce:1}]`. When we arrive at nonce 1, the candidates will be nonce 2 and 3. Then we reevaluate for 2, which again has 3 as a candidate, so we reevaluate those, all of them pass. But on the first level of the recursion, nonce 3 is still in the array of candidates, leading to it's reevaluation once more. 
Can you think of a way how we can prevent that?

packages/sequencer/src/mempool/private/PrivateMempool.ts
```diff
@@ -50,8 +80,98 @@ export class PrivateMempool extends SequencerModule implements Mempool {
...
+            }
+          }
+        } else {
+          console.log(
```
rpanic: console.log -> log

packages/sequencer/test/integration/Mempool.test.ts
```diff
@@ -0,0 +1,379 @@
...
+        "EKEiWucbahaja6beCxVY72ayknTWWwT6WQHHun9bZdkcdboAA8yS"
+      );
+      const user2PublicKey = user2PrivateKey.toPublicKey();
+
+      const user3PrivateKey = PrivateKey.fromBase58(
+        "EKDvi75A2GN1vBN5Tyxfzf1H7rHRvn7abKCTBzz6Vcv6LjbMWEvJ"
+      );
+      const user3PublicKey = user3PrivateKey.toPublicKey();
+
+      await mempool.add(
```
rpanic: I think we can potentially create a helper function that makes this code cleaner?

packages/sequencer/test/settlement/Settlement.test.ts
```diff
@@ -167,8 +167,8 @@ describe("settlement contracts", () => {
 
       await mempool.add(tx);
     }
-    txs.forEach((tx) => {
-      mempool.add(tx);
+    txs.forEach(async (tx) => {
```
rpanic: mapSequential, those promises won't be awaited in this case. But good catch!

packages/sequencer/src/mempool/private/PrivateMempool.ts
ejMina226: As discussed, instead of using recursion we have instead opted for a dynamic list.

packages/sequencer/src/mempool/private/PrivateMempool.ts
ejMina226: I've extracted the function outside of the `getTexs()` and added some comments.

packages/sequencer/src/mempool/private/PrivateMempool.ts
```diff
@@ -50,11 +81,116 @@ export class PrivateMempool extends SequencerModule implements Mempool {
...
+          // eslint-disable-next-line @typescript-eslint/no-loop-func
+          stateTransitions.forEach((st) => {
+            Object.values(skippedTransactions).forEach((value) => {
+              if (value.paths.some((x) => x.equals(st.path))) {
+                transactions.push(value.transaction);
+              }
+            });
+          });
+
+          // eslint-disable-next-line no-param-reassign
```
rpanic: There is a reason for this eslint rules :) So imo we should create a duplicate and return it from the function call. 
So the signature would look like `(pendingTxs: Tx[]) => Tx[]` where the result are the accepted/sorted transactions. Because in the end we also don't care about the skipped transactions after this functions returns, so there is no reason to leak it out of this scoped

packages/sequencer/src/mempool/private/PrivateMempool.ts
```diff
@@ -50,11 +81,116 @@ export class PrivateMempool extends SequencerModule implements Mempool {
...
+          transactions = transactions.filter(
```
rpanic: We have utils functions called `distinctByPredicate` or `distinctByString` in the sequencer utils, so we might use those to make that part more readable. We could also think about using `Set`, which automatically deduplicates stuff, not sure if that works with complex objects though, you'd have to check

packages/sequencer/src/mempool/private/PrivateMempool.ts
rpanic: Then, as the second step, we remove all the returned Txs from the transactionstorage (that happens implicitly by including them in a block and storing that relationship in the DB

packages/sequencer/src/mempool/private/PrivateMempool.ts
```diff
@@ -50,11 +81,116 @@ export class PrivateMempool extends SequencerModule implements Mempool {
...
+      executionContext.setup(contextInputs);
+
+      const signedTransaction = tx.toProtocolTransaction();
+      // eslint-disable-next-line no-await-in-loop
+      await this.accountStateHook.onTransaction({
+        networkState: networkState,
+        transaction: signedTransaction.transaction,
+        signature: signedTransaction.signature,
+      });
+      delete transactions[index];
```
rpanic: Also this doesn't work as you'd expect, see here https://stackoverflow.com/a/500617

ejMina226: I chose this delete as it was one of the few that updated the variable in place. However, thinking about it now, I'm not sure we even need to delete for the transactions/queue as once we iterate over it it doesn't matter anyway.

ejMina226: The skippedTransactions will still need to be deleted, though.

PR #208, opened 2024-10-14
Reorganized VKTree services and implemented startup consumption of sequencer arguments for workers

### LINKED ISSUE #207: Allow workers to consume runtime values at startup
Currently, the only "input" workers get for circuit compilation is the source code itself.
For advances usecases, for example, sideloading runtime methods or provably deploying the bridging contract from the settlement contract, dynamic inputs are needed that need to be baked into the circuits at compile time.

These values typically are collected by the sequencer and require a large amount of computation to be computed. Therefore, workers should be able to receive those arguments at startup from the sequencer.

#### Protocol

To solve this, there will be a message queue that reserves a dedicated queue for workers to register themselves to the sequencer and receive arguments in the process. For that, as soon as the sequencer is up and running, the sequencer dispatches a task with the arguments to that queue. As soon as any worker gets spun up, they listen for and consume that task where the arguments are sent as payload. After reading that task, the worker "completes" the task - i.e. sends a response back to the message queue, which the sequencer receives. When the sequencer sees that acknowledgement, it dispatches the same task again for any subsequently spun-up worker to consume and the protocol to start again

rpanic:
Closes #207 

Description in the issue

Enhanced the FlowTaskWorker to accept three different types of tasks:
`UnpreparingTask`: Tasks that don't implement a prepare() method, therefore they don't have to wait for registration to get initialized - important because we need the CircuitCompilerTask before we have working worker registration
`AbstractStartupTask`: Tasks that are used for startup of workers: Mainly the WorkerRegistrationTask
`Task`: All normal tasks that wait for registration to be completed before preparing and registering for tasks


ejMina226 (approved):
Looks good.

ejMina226:
@rpanic From a high-level, how did you get around the issue of the worker processes not having access to the VKTreeService?

rpanic:
@ejMina226 
Two-fold: The first thing I did was instead of accessing the VKTreeService in the BlockProver, which is executed by a task, I add the VK and the tree witness as a task payload and through that, I use them in the BlockProver as a traditional program arguments / witness.
The second thing is how the blockprover has access to the tree root, which has to be baked into the circuit at compile time. This is done via a extra step that happens when the worker gets started, which is the "worker registration". I outlined that in more detail in the linked issue

PR #209, opened 2024-10-16
Make Block Size Configurable

ejMina226:
Add a configurable maximum block size that the mempool respects


packages/sequencer/src/protocol/production/sequencing/BlockProducerModule.ts
```diff
@@ -144,7 +149,7 @@ export class BlockProducerModule extends SequencerModule<BlockConfig> {
     txs: PendingTransaction[];
     metadata: BlockWithResult;
   }> {
-    const txs = await this.mempool.getTxs();
+    const txs = (await this.mempool.getTxs()).slice(0, this.maximumBlockSize());
```
rpanic: I would like to see this handled in a more eager fashion by passing the maximum size to getTxs() and then returning there as soon as that size is hit (instead of continuing to simulate and run everything)

ejMina226: OK. I'll change `getTx` to take an argument and do the filtering there.

PR #212, opened 2024-10-25
Fixed issue where Mempool endpoints would invoke tx ordering

rpanic:
Also: Made findTransaction() in Inmemory and Prisma TxStorage behave the same


ejMina226 (approved):
LGTM. Why do we not want tx ordering here?

rpanic:
Because it's irrelevant to the result of the endpoint (the tx is either included or not, the order doesn't matter there), but consumes a lot of resources computing the result, and that is a very apparent DOS vector for any attacker

PR #214, opened 2024-10-25
Fixed execution context not cleared after getTxs invocation

rpanic:


ejMina226 (approved):
LGTM. What happens if the execution context isn't cleared? Do the effects from the txs start accumulating?

rpanic:
Yes, I got a bug report from a user that discovered that, it was in conjunction with the MempoolResolver API endpoint though, but the result was a polluted context that leaked incorrect data into the block production somehow

PR #222, opened 2024-11-07
Refactor STProver to Accept Merkle Witness as Argument

ejMina226:
Previously, there was a getWitnessProvider call in the StateTransitionProver, which was misleading as it suggested the possibility of being able to fetch any witness at any time. This is not the case, and there is just one witness for each StateTransition available.


packages/protocol/src/prover/statetransition/StateTransitionProver.ts
```diff
@@ -132,14 +131,6 @@ export class StateTransitionProverProgrammable extends ZkProgrammable<
     ];
   }
 
-  private get witnessProvider(): StateTransitionWitnessProvider {
```
rpanic, line 135: Lets also remove all the classes that are unused now because of that, they shouldn't be used anywhere else now

packages/protocol/src/prover/statetransition/StateTransitionProver.ts
```diff
@@ -187,12 +185,9 @@ export class StateTransitionProverProgrammable extends ZkProgrammable<
     state: StateTransitionProverExecutionState,
     transition: ProvableStateTransition,
     type: ProvableStateTransitionType,
+    witness: RollupMerkleTreeWitness,
```
rpanic: Let's call that merkleWitness here as well, outwise it's confusing with the Provable.witness

packages/sequencer/src/protocol/production/tasks/StateTransitionTask.ts
```diff
@@ -68,6 +68,7 @@ export class StateTransitionTask
     witnessProviderReference.setWitnessProvider(witnessProvider);
```
rpanic: We should also get rid of all the witness provider stuff here

PR #225, opened 2024-11-09
Enabled sourcemaps

rpanic:
This PR enables TS sourcemaps for better error messages and code lookups.
Also, this enables out-of-package debugging in IntelliJ


ejMina226 (approved):
LGTM

PR #233, opened 2024-11-22
Created transactions on Database to do atomic writes for blocks, results and batches

### LINKED ISSUE #231: Atomic Block Production
Currently, block production isn't atomic and consistent when it comes to DB writes. If the block production fails in the middle of production or between production and result generation, inconsistent states can occur.
To fix this, two things should be implemented:
- Make result generation a seperate step in the production pipeline and enable it resume operation in case the sequencer crashed after production and before result generation
- Create a DB transaction pattern that allows us to encapsulate all of the DB writes done during block production in a single transaction

rpanic:
Closes #231

### DIFF
--- a/packages/persistance/src/PrismaDatabaseConnection.ts
+++ b/packages/persistance/src/PrismaDatabaseConnection.ts
@@ -137,4 +137,8 @@ export class PrismaDatabaseConnection
   public async close() {
     await this.prismaClient.$disconnect();
   }
+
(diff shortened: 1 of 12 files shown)


maht0rz:
Can we add a test that demonstrates the issue that led to this fix?

rpanic:
Test added, pls re-review @maht0rz

maht0rz (approved):

PR #243, opened 2024-12-13
Fix for Mempool and BP using different State Services

ejMina226:
A community member reported seeing an issue where a tx was stuck in pending, i.e. never included in a block. This was because when a block is produced it checks to see if a tx is valid, which involves checking the nonce (amongst other things). A previous tx had already been sent from the same sender, but this was not being noticed by the BP and so expected the tx in the mempool to have nonce 0, rather the the correct nonce of 1, which it in fact had. Consequently, the tx was never picked up and just sat there in the mempool.

### DIFF
--- a/packages/sequencer/test/integration/BlockProduction.test.ts
+++ b/packages/sequencer/test/integration/BlockProduction.test.ts
@@ -337,6 +337,107 @@ describe("block production", () => {
     expect(newState).toBeUndefined();
   }, 30_000);
 
+  it("should produce txs in non-consecutive blocks", async () => {
+    log.setLevel("TRACE");
+
+    const privateKey = PrivateKey.random();
+    const publicKey = privateKey.toPublicKey();
+
+    const privateKey2 = PrivateKey.random();
+    const publicKey2 = privateKey2.toPublicKey();
+
+    await mempool.add(
+      createTransaction({
+        runtime,
+        method: ["Balance", "setBalanceIf"],
+        privateKey,
+        args: [publicKey, UInt64.from(100), Bool(true)],
+        nonce: 0,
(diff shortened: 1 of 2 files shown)


packages/sequencer/test/integration/BlockProduction.test.ts
```diff
@@ -337,6 +337,107 @@ describe("block production", () => {
     expect(newState).toBeUndefined();
   }, 30_000);
 
+  it("should produce txs in non-consecutive blocks", async () => {
```
rpanic, line 340: @ejMina226 What is the unique thing of this test that isn't covered in the others? Can this be modeled using the multiple blocks test in line 471(ish)?

ejMina226: I wanted the txs to appear in non-consecutive blocks.

ejMina226: The `should produce multiple blocks with multiple batches with multiple transactions` has consecutive blocks of txs, I think.

ejMina226: Happy to remove if you think it overkill.

rpanic: What do you mean by non-consecutive? If I understand it correctly, you have 1 tx per block here

ejMina226: There are two different senders. I wanted the first sender to have their txs in non-consecutive blocks. The second sender is just to ensure the interim blocks are non-empty.

rpanic (approved):

PR #245, opened 2024-12-13
Record resolving order in the sequencer and start modules based on that

### LINKED ISSUE #240: Record resolving order in the sequencer and start modules based on that
Currently, the order that sequencer modules are resolved can be different in which they are started, which can lead to problems on dependencies not being initialized properly.
We can fix this by recording the order that modules are resolved and starting them based on that.

ejMina226:
Presently, the sequencer uses `tsyringe` to resolve modules (and their dependencies) and then starts them. However, this can be problematic as although `tsyringe` may resolve dependencies, it doesn't actually start them. For example, a database may be created, but the connection strings, etc, won't be constructed until it's started, and this may cause an error if a module that relies on it is started first. The way to resolve this is ensure that we start modules based on the order they were resolved.

### DIFF
--- a/packages/sequencer/src/sequencer/executor/Sequencer.ts
+++ b/packages/sequencer/src/sequencer/executor/Sequencer.ts
@@ -66,6 +66,14 @@ export class Sequencer<Modules extends SequencerModulesRecord>
    * modules to start each
    */
   public async start() {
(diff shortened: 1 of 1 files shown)


rpanic (approved):
Lgtm

PR #249, opened 2024-12-19
feature/remove-zkProgrammable

ejMina226:
The latest version of o1js includes the ability to mock a zkProgram. This was the original motivation behind the zkProgrammable abstract class and so it can be removed. This PR includes the following:

1. Upgrade o1js to version 2.
2. Fix the types that have changed because of this.
3. Remove `zkProgrammable` and `WithZkProgrammable` and use `zkProgramFactory` and `WithZkProgram`


ejMina226:
Closing this PR as we need to upgrade to a later npm version and this requires changes to the codebase. There is already a branch with these changes in, so will rebase off that. See #256

Issue #199, opened 2024-09-11
Add protection against missing awaits in runtime code


rpanic:
A common pitfall that users encounter is to forget to add `await` to `state.set()` calls. This mostly leads to ambigous errors and takes up a lot of time to debug - especially when doing community support.
We can approach this problem from two angles:
1. Add better eslint coverage for missing awaits in the starter-kit
2. Build in a framework-based solution that enforces the linearity of any runtime or protocol code. This can be done pretty easily by doing a analyzing step of all user-built code where we keep track of a global context (and how much it has been entered) for all state calls and add a slight (internally awaited) delay to force the js evaluation loop to continue elsewhere to catch missing awaits

PR #253, opened 2025-01-10
Mitigate issues with missing awaits in state method calls

joaosreis:
This PR creates a queue-like mechanism for the state methods with a method that can be awaited until all the queued promises are resolved. This method is then called after a runtime method executes, which makes is so that these state operations are executed in the proper runtime moment (and in the order they are queued) even if the user misses to await them.
This helps mitigate the issue of missing awaits to state methods (#199), ensuring these are properly executed if they are runtime methods top level calls, but not if the runtime method calls some other method with state calls and this method is not awaited, because then the runtime method execution can finish before the other method had the chance to queue the operations. An example of such behavior would be:
```typescript
@runtimeMethod()
  public async addBalance(
    tokenId: TokenId,
    address: PublicKey,
    amount: Balance
  ): Promise<void> {
    const circulatingSupply = await this.circulatingSupply.get();
    const newCirculatingSupply = Balance.from(circulatingSupply.value).add(
      amount
    );
    assert(
      newCirculatingSupply.lessThanOrEqual(this.config.totalSupply),
      "Circulating supply would be higher than total supply"
    );
    this.circulatingSupply.set(newCirculatingSupply);
    this.mint(tokenId, address, amount); // missing await in a method that calls state methods
}
```
However, if we deconstruct the `mint` method into
```typescript
const key = new BalancesKey({ tokenId, address });
const balanceOption = await this.balances.get(key);
const balance = Balance.Unsafe.fromField(balanceOption.value.value);
const newBalance = balance.add(amount);
this.balances.set(key, newBalance);
```
this code will now work despite the missing `await`s in the state set calls.

I also tested adding a 50ms delay timeout to the queue `onCompleted` method in an attempt to help these other missing `await`s catchup, but ended up not adding it as I felt that it might not work if the complexity of the runtime methods increases as well because this doesn't ensure the linearity of these calls.


rpanic:
Not really the approach we were going for, closing in favor of a future implementation

PR #259, opened 2025-01-20
Refactor injection token definitions for dynamic keys

### LINKED ISSUE #39: Starting the app-chain client in a UI build / not during dev results into dependency errors with transaction/block hooks
**Reported here:**
https://discord.com/channels/484437221055922177/1319673057849380995/1319673057849380995

**Full report:**
Hi, I am having this 2 errors when trying to `.start()` the client inside the Next.js (client-side).

```
Can't inject hook under the underlying hook token: Alias not found in mapping
```

```
Uncaught (in promise) Error: Cannot inject the dependency "n" at position #2 of "eX" constructor. Reason:
    Attempted to resolve unregistered dependency token: "ProvableTransactionHook"

or:

    Attempted to resolve unregistered dependency token: "ProvableBlockHook"
```

The strange thing is that this second error only happens after build and start, not in development.

I was able to fix the second error by doing:
```typescript
import { container } from "tsyringe";
import { ProvableBlockHook, ProvableBlockHook } from "@proto-kit/protocol";

container.registerInstance("ProvableBlockHook", ProvableBlockHook);
container.registerInstance("ProvableBlockHook", ProvableBlockHook);
```

However I don't think it should be permanent. So please let me know if there is any way to fix or debug.

Note I am using React 18 but Next.js 15.

joaosreis:
Fixes proto-kit/starter-kit#39, which was caused by some update to Next.js/Webpack on Next.js v15 that would make class names mangled on production builds, rendering in failed dependency injections that relied on the class names. This PR addresses this by setting the keys of the record with the injection tokens to be set to the dynamic value of the class names.

### DIFF
--- a/packages/protocol/src/protocol/Protocol.ts
+++ b/packages/protocol/src/protocol/Protocol.ts
@@ -27,6 +27,12 @@ import { ProvableTransactionHook } from "./ProvableTransactionHook";
 import { ProtocolEnvironment } from "./ProtocolEnvironment";
 import { ProvableBlockHook } from "./ProvableBlockHook";
 
+/**
+ * This is a mapping of abstract classes to their respective injection tokens.
+ * Keys are the abstract classes names, which need to be set dynamically
+ * and can't be hardcoded since producing optimized builds may mangle the
+ * class names, making them different from the ones in the source code.
+ */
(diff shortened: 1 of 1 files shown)


rpanic:
Can you comment/document this issue somewhere fitting in the code? So that we don't stumble over this very hidden issue in the future

PR #280, opened 2025-02-13
Add Events to Protocol Hooks

### LINKED ISSUE #250: Runtime Events and Protocol Hooks
Currently, you can technically emit runtime events inside of protocol hooks via runtime calls.
However, these events won't be picked up anywhere. We should consolidate that primitive to also extend to protocol hooks

ejMina226:


packages/persistance/src/services/prisma/mappers/EventMapper.ts
```diff
@@ -7,42 +7,76 @@ import { ObjectMapper } from "../../../ObjectMapper";
...
+    >
 {
-  public mapIn(input: Prisma.JsonObject): { eventName: string; data: Field[] } {
-    if (input === undefined) return { eventName: "", data: [] };
+  public mapIn(input: Prisma.JsonObject): {
+    eventName: string;
+    data: Field[];
+    source: "afterTxHook" | "beforeTxHook" | "runtime";
+  } {
+    if (input === undefined)
```
rpanic: This check is not needed, the input cannot be undefined

packages/persistance/src/services/prisma/mappers/EventMapper.ts
```diff
@@ -7,42 +7,76 @@ import { ObjectMapper } from "../../../ObjectMapper";
...
+    );
+  }
 }
 
 @singleton()
 export class EventArrayMapper
   implements
     ObjectMapper<
-      { eventName: string; data: Field[] }[],
+      {
```
rpanic: I think now that this type has become so complex, we should extract its definition to a "type EventData" at the beginning of this file

packages/sequencer/src/protocol/production/sequencing/TransactionExecutionService.ts
```diff
@@ -394,7 +406,7 @@ export class TransactionExecutionService {
         statusMessage: runtimeResult.statusMessage,
 
         stateTransitions,
-        events,
+        events: beforeHookEvents.concat(afterHookEvents, runtimeResultEvents),
```
rpanic: We should keep this in the order they've been emitted

PR #281, opened 2025-02-13
Added static dependency factories

rpanic:
Added static dependencyfactories
Fix resolving issue by making IncomingMessagesService dependency static

Merge into #272


maht0rz (approved):

maht0rz:
Only issue i see with this is that if deps are static, they won't have access to module configuration, if that's something we are okay with then lgtm

PR #283, opened 2025-02-14
Model STs in Persistence

### LINKED ISSUE #261: Properly model StateTransition and -Batches in peristence
Currently STs and corresponding STBatches are modelled as JSON inside postgres.
We ideally want to minimize usage of JSON fields, since that 1) doesn't allow querying over them and 2) locks us more and more into postgres

Therefore, a relational model should be built to save STs properly

ejMina226:
Instead of storing the STs and STBatches as json in a table, which makes them hard to query, we now create separate tables to store the information. This also requires changes to the mappers.


packages/persistance/test/stMapper.test.ts
```diff
@@ -0,0 +1,45 @@
...
+      from: { isSome: false, value: [], isForcedSome: false },
+      to: { isSome: false, value: [], isForcedSome: false },
+    },
+  ])("MapOut to MapIn", (input) => {
+    const untypedTransition = UntypedStateTransition.fromJSON(input);
+    const stMapper = new StateTransitionMapper();
+    const result = stMapper.mapIn(stMapper.mapOut(untypedTransition)).toJSON();
+    expect(result).toEqual(input);
+  });
+  it.each([
```
rpanic, line 26: Empty line missing

packages/persistance/src/services/prisma/mappers/StateTransitionMapper.ts
```diff
@@ -3,78 +3,96 @@ import {
...
-  public mapIn(input: Prisma.JsonObject): UntypedStateTransition {
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
-    return UntypedStateTransition.fromJSON(input as any);
-  }
-
-  public mapOut(input: UntypedStateTransition): Prisma.JsonObject {
-    return input.toJSON();
-  }
-}
+export type STBatchArrayMapOut1 = Omit<
```
rpanic: Can we give those types better names? :D

packages/persistance/src/services/prisma/mappers/StateTransitionMapper.ts
```diff
@@ -3,78 +3,96 @@ import {
...
+  DBStateTransitionBatch,
+  "txExecutionResultId" | "id" | "blockId" | "blockResultId"
+>;
+export type STBatchArrayMapOut2 = Omit<DBStateTransition, "batchId" | "id">[];
```
rpanic: Generally, I like this pattern of re-defining the types that are derived from the Prisma types. I see that there are a lot of (by now too complicated) types that do this, can we extract them into those reusable defintions where it makes sense? Might make stuff more readable, especially at places where we stich them together

rpanic:
Another very important thing: We need to add the new tables to the PrismaDatabase's `.pruneDatabase()` tables @ejMina226

PR #288, opened 2025-02-27
Don't emit STs when accessing STs outside provable code

### LINKED ISSUE #242: Don't emit STs when accessing STs outside provable code
Currently, when users use Provable.witness() or Provable.asProver blocks and access the state within, STs will be emitted, which aren't enforced provably though. 
We should add in a check that returns the state, but doesn't emit STs in the process.
Additionally, .set calls should be disallowed in witness blocks

ejMina226:


packages/protocol/src/state/State.ts
```diff
@@ -36,6 +36,44 @@ export class WithStateServiceProvider {
   }
 }
 
+@singleton()
+export class IsInWitnessBlockContext {
+  public isInWitnessBlock: number = 0;
+}
+
+const newAsyncWitnessFunction = (
```
rpanic: Lets type these functions properly (I used any to prototype it)

packages/protocol/src/state/State.ts
```diff
@@ -36,6 +36,44 @@ export class WithStateServiceProvider {
...
+    const context = container.resolve(IsInWitnessBlockContext);
+    context.isInWitnessBlock += 1;
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    const ret = await originalFuncDef(e, f);
+    context.isInWitnessBlock -= 1;
+    return ret;
+  };
+};
+
+const newWitnessFunction = (originalFuncDef: any) => {
```
rpanic: call this proxyWitnessFunction?

packages/protocol/src/state/State.ts
```diff
@@ -36,6 +36,44 @@ export class WithStateServiceProvider {
...
+    const context = container.resolve(IsInWitnessBlockContext);
+    context.isInWitnessBlock += 1;
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    const ret = originalFuncDef(e, f);
+    context.isInWitnessBlock -= 1;
+    return ret;
+  };
+};
+
+const originalWitnessAsync = Provable.witnessAsync;
```
rpanic: Can be condensed into single line

packages/protocol/src/state/State.ts
```diff
@@ -36,6 +36,44 @@ export class WithStateServiceProvider {
   }
 }
 
+@singleton()
+export class IsInWitnessBlockContext {
+  public isInWitnessBlock: number = 0;
```
rpanic: this naming is a relict from when it was a boolean during prototyping, lets change this to `witnessBlockDepth` and add a `isInWitnessBlock` getter

PR #291, opened 2025-03-05
Make it easier to access parent dependency container inside modules 

### LINKED ISSUE #276: Make it easier to access parent dependency container inside modules
Currently, in our codebase, we have a lot of spots like this:
```typescript
public constructor(
  @inject("Sequencer")
  private readonly sequencer: Sequencer<SequencerModulesRecord>
) {}
```

so we can then later in some method of the module do

```
this.sequencer.dependencyContainer.resolve("Foo");
```

Therefore, I propose that every ModuleContainer self-injects itself in its own dependency container, so that modules can have a simpler syntax:
```typescript
public constructor(
  @inject("ParentContainer")
  private readonly parent: ParentModuleContainer
) {}
  
```

ejMina226:
A slicker approach to accessing the parent container of a module.


rpanic:
@ejMina226 Can we also change all other existing references to "Sequencer", "Protocol" and "Runtime" to "ParentContainer"?

rpanic (approved):

PR #294, opened 2025-03-11
Solution to Failed TX Hooks

### LINKED ISSUE #217: Figure out how to deal with transaction where the tx hooks fail
Currently, if a transaction is included in a block but one of the transaction hooks fails, we skip the transaction in that block, but it won't get removed from the mempool. This leads to the transaction being applied (and failing) again and again, for some users that is unwanted behaviour.

We need to figure out how to deal with that, especially also in the context that transaction might become valid in the future. For example, the nonce might be too large for a particular account, but if another transaction with the right nonce comes along, the transaction might become valid again. 
For this case, we might keep them in the mempool for the future, but we need to distinguish them from transactions that will never be valid, or ones that we want to throw out regardless.

### LINKED ISSUE #321: Handling of failed tx hooks for incoming messages
Currently, if a tx hook fails for incoming messages, this will lead to a unrecoverable state because the message hash will not be added to the honored messages list.
Therefore the L1 action state requirement will never be met.

To fix this, we have to do three things:
- Assert tx hooks failing only for non-messages.
- For messages with failing tx hooks, append the message hash but throw away everything else.
- Adapt the block production to account for both previous proofs

ejMina226:
Closes #321 

Currently, when a tx hook fails the tx is not removed from the mempool. In some cases this is what's needed. However, in others, like when a nonce is too small, this is a problem as the tx will persist in the memppool indefinitely, occupying resources. The solution is to incorporate a new check into the txHook abstract class, called `removeTransactionWhen`, which when satisfied causes the tx to be removed from the mempool.


packages/sequencer/src/mempool/private/PrivateMempool.ts
```diff
@@ -175,6 +175,23 @@ export class PrivateMempool
       executionContext.setup(contextInputs);
 
       const signedTransaction = tx.toProtocolTransaction();
+
```
rpanic, line 197: What I'd expect here is that this only gets evaluated when the transaction is not included in a block, i.e. when the hook itself fails. Otherwise, we shouldn't evaluate `removeTransactionWhen` at all

rpanic (changes requested):
The comment above, and also let's get the CI working, then i'd say its gtg

packages/sequencer/src/mempool/private/PrivateMempool.ts
```diff
@@ -110,16 +110,22 @@ export class PrivateMempool
     return result?.result.afterNetworkState;
   }
 
+  public async removeTxs(included: string[], dropped: string[]) {
+    await this.transactionStorage.removeTx(included, "included");
```
kadirchan, line 114: `if (type === "dropped")` check in `removeTx()` method makes this call redundant, we can refactor `removeTxs()` if there is a need of dropping for included ones or remove that call.
Mentioned [removeTx()](https://github.com/proto-kit/framework/blob/02973714a909bd693eced8200a4ae869d79d6a2d/packages/persistance/src/services/prisma/PrismaTransactionStorage.ts#L38-L39)

packages/sequencer/src/mempool/private/PrivateMempool.ts
```diff
@@ -128,7 +134,7 @@ export class PrivateMempool
           networkState,
           limit
         )
-      : txs;
+      : txs.slice(0, limit);
```
kadirchan, line 137: I think it would be better if we add limit parameter to getPendingUserTransactions method with db level limiting instead and slicing after getting all transactions. This could be a performance problem as txs grows

packages/sequencer/src/storage/inmemory/InMemoryTransactionStorage.ts
```diff
@@ -19,6 +19,13 @@ export class InMemoryTransactionStorage implements TransactionStorage {
     @inject("BatchStorage") private readonly batchStorage: InMemoryBatchStorage
   ) {}
 
+  public async removeTx(hashes: string[]) {
+    this.queue = this.queue.filter((tx) => {
+      const hash = tx.hash().toString();
+      return !hashes.includes(hash);
+    });
+  }
```
kadirchan, line 28: we can reduce lookup complexity in here 

```suggestion

  public async removeTx(hashes: string[]) {
    const hashSet = new Set(hashes)
    this.queue = this.queue.filter((tx) => {
      const hash = tx.hash().toString();
      return !hashSet.has(hash);
    });
  }
```

kadirchan: also would be better caching that set

packages/protocol/src/hooks/AccountStateHook.ts
```diff
@@ -54,4 +54,19 @@ export class AccountStateHook extends ProvableTransactionHook {
   public async afterTransaction() {
     noop();
   }
+
+  // Under these conditions we want the tx removed from the mempool.
+  public async removeTransactionWhen({
```
kadirchan, line 59: It might be a silly question, but in the current version I couldn't be sure whether the transaction gets removed from the mempool when there isn't enough gas fee. Also, an option could be added for after a certain timeout period.

kadirchan (commented):
reviewed with my limited knowledge about codebase but looks good overall. Added couple things

packages/sequencer/src/mempool/private/PrivateMempool.ts
rpanic: Absolutely - that's covered in #330

packages/sequencer/src/mempool/private/PrivateMempool.ts
rpanic: While what you're saying is true - its only the case for the PrismaTransactionStorage, not for the InMemoryTransactionStorage. This is because they are implemented differently, and for the inmemory version, we will need to remove them explicitely (while for prisma, the DB schema takes care of it via foreign key relationships)

packages/protocol/src/hooks/AccountStateHook.ts
rpanic: Added it, thanks for spotting this!

PR #295, opened 2025-03-26
Implemented smart hashing of batch updates

rpanic:
This PR will only become relevant in case of implementation of #191 
The speedup on a normal sparse tree is 0.6% on a height of 256. 
With the LMT, since the leafs will be placed next to each other, the speedup should be more significant


rpanic:
Closed in favor of #305

PR #323, opened 2025-11-04
Upgrade Framework o1js Version

kadirchan:
This PR upgrades the o1js version used in framework to latest `2.10.0`


packages/common/src/zkProgrammable/provableMethod.ts
```diff
@@ -49,8 +50,13 @@ export function toProver(
 
       // TODO: provide undefined if public input is not used
       publicInput: isFirstParameterPublicInput ? args[0] : undefined,
-      publicOutput,
-
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+      publicOutput:
```
rpanic: Would you agree that `publicOutput?.publicOutput` would be the same semantically?

packages/api/package.json
```diff
@@ -48,7 +48,7 @@
     "@proto-kit/module": "*",
     "@proto-kit/protocol": "*",
     "@proto-kit/sequencer": "*",
-    "o1js": "^1.6.0",
+    "o1js": "^2.1.0",
```
rpanic: Current o1js version is 2.10.0, is this a typo or any particular reason you didn't go to 2.10 directly?

packages/protocol/src/prover/block/BlockProver.ts
```diff
@@ -932,11 +932,13 @@ export class BlockProverProgrammable extends ZkProgrammable<
             runtimeProof: DynamicRuntimeProof,
             executionData: BlockProverSingleTransactionExecutionData
           ) {
-            return await proveTransaction(
-              publicInput,
-              runtimeProof,
-              executionData
-            );
+            return {
+              publicOutput: await proveTransaction(
```
rpanic, line 936: Omg I hate that o1js made this the API, so verbose

rpanic (changes requested):
Two questions, failing test should be easy to fix, otherwise good job!

packages/common/src/zkProgrammable/provableMethod.ts
kadirchan: It would but we're getting {} instead of null or undefined as publicOutput in [const publicOutput = await Reflect.apply(simulatedMethod, this, args);](https://github.com/proto-kit/framework/blob/a1cd3e195512b78fdd660950369358829d49b1eb/packages/common/src/zkProgrammable/provableMethod.ts#L46-L47) thus this approach gives `Property 'publicOutput' does not exist on type '{}'.` if we don't check explicitly.

packages/api/package.json
kadirchan: Just in case, I added the same version as the updated mina fungible token, but I thought it wouldn't be a problem since it already uses the latest version on first installation anyway. I fixed it.

packages/common/src/zkProgrammable/provableMethod.ts
rpanic: Okay got it. Do we have cases where the type is only a single "publicOutput" level so to speak? Because if so, we should probably find out where and make it the same everywhere

rpanic (approved):

PR #326, opened 2025-11-05
Added warning message for missing BatchProducerModule

rpanic:


kadirchan (approved):
I think we should document configuration settings in a way or give it through maybe jsdoc to prevent missing modules. Its confusing and easy to make mistake

PR #327, opened 2025-11-06
Feat/query module

### LINKED ISSUE #204: Enhance client-side query API
Users are missing two features of the query API:
- Have a function that awaits the inclusion of some transaction
- Query data of specific, recent blocks

Currently, both of those functionalities are done via raw graphql queries in the starter kit, we should package them more nicely

4ytuncs:
Closes #204 

This PR is intended to implement the query features mentioned in issue #204 as a module.

#### Core Implementation
- Added `BlockExplorer` interface defining `waitTxInclusion()` and `getBlock()` methods F(338ca4e595a70338ae9e3b513918b8de387c1555)
- Implemented `GraphqlBlockExplorer` in `@proto-kit/sdk` for client-side remote queries
- Implemented `InMemoryBlockExplorer` in `@proto-kit/sequencer` for server-side local queries (Final form of those two are at 92fa991359f5d6fb7b67ade5ef5963352366be3e)

#### Features
- `waitTxInclusion()`: Poll transaction status until included in a block
  - Returns transaction state: INCLUDED, PENDING, or UNKNOWN
  - Configurable polling interval and attempt limits
- `getBlock()`: Retrieve block data by hash or height
  - Returns block model with transactions and metadata
  - Defined GraphQL fragment for efficient query composition (92fa991359f5d6fb7b67ade5ef5963352366be3e)

#### Tests
- Added comprehensive test cases for both explorer implementations
- Tests cover transaction inclusion polling and block retrieval


packages/sdk/src/client/ClientAppChain.ts
```diff
@@ -212,6 +217,8 @@ export class ClientAppChain<
 
     const network = new NetworkStateQuery(networkStateTransportModule);
 
+    const explorer = this.container.resolve<BlockExplorer>("BlockExplorer");
```
rpanic: Generally, I'd prefer this to adapt the structure of the other existing query modules. So to have a transport module implementing the actual interface to the API (either graphql or the stores directly) and then a `...Query` service that implements the remaining semantics

packages/sdk/src/graphql/GraphqlExplorer.ts
```diff
@@ -0,0 +1,109 @@
...
+      await sleep(interval);
+    }
+  }
+
+  public async getBlock(
+    blockHash: string,
+    blockHeight: number
+  ): Promise<unknown> {
+    const query = gql`
+      ${BATCH_TRANSACTION_MODEL_FRAGMENT}
```
rpanic: What benefits does the fragment have here? I'm not super familiar with their functionalities

packages/sdk/src/graphql/GraphqlExplorer.ts
```diff
@@ -0,0 +1,109 @@
...
+    const query = gql`
+      query transactionState($hash: String!) {
+        transactionState(hash: $hash)
+      }
+    `;
+
+    let remainingAttempts = attempts;
+
+    // eslint-disable-next-line no-constant-condition
+    while (true) {
```
rpanic: This while loop along with checks/etc is the thing that I'd want implemented in the BlockExplorerQuery, and then the graphql calls inside of the transport modules

packages/sdk/src/graphql/GraphqlExplorer.ts
```diff
@@ -0,0 +1,109 @@
...
+      const status = queryResult.data?.transactionState as
+        | string
+        | undefined;
+
+      if (status === "INCLUDED") {
+        return { transactionState: InclusionStatus.INCLUDED };
+      }
+
+      if (status === "PENDING") {
+        return { transactionState: InclusionStatus.PENDING };
```
rpanic: If we `waitTxInclusion` I'd assume that don't return with state `Pending` but instead wait until that Pending turns into Included.

packages/sdk/src/graphql/GraphqlExplorer.ts
```diff
@@ -0,0 +1,109 @@
...
+        }
+      }
+    `;
+
+    const queryResult = await this.graphqlClient.client
+      .query(query, { hash: blockHash, height: blockHeight })
+      .toPromise();
+
+    if (queryResult.error) {
+      throw new Error("Error fetching block!");
```
rpanic: Lets relay the exact error as well, so prefix it like this, but then append the gql error we received

packages/sdk/src/query/InMemoryBlockExplorer.ts
```diff
@@ -0,0 +1,75 @@
...
+    while (true) {
+      // eslint-disable-next-line no-await-in-loop
+      const dbTx = await this.transactionStorage.findTransaction(txHash);
+
+      if (dbTx?.block !== undefined) {
+        return { transactionState: InclusionStatus.INCLUDED };
+      }
+
+      if (dbTx !== undefined) {
+        return { transactionState: InclusionStatus.PENDING };
```
rpanic: Same as in the Gql implementation

packages/sequencer/src/helpers/query/BlockExplorer.ts
```diff
@@ -0,0 +1,12 @@
+export interface BlockExplorer {
+  waitTxInclusion(
+    txHash: string,
+    interval?: number,
```
rpanic: Lets add tsdocs here describing what unit `interval` is

packages/stack/test/graphql/graphql-server.ts
```diff
@@ -210,6 +213,8 @@ export async function startGraphqlServer() {
     },
   });
 
+  container.registerInstance("BlockExplorer", InMemoryBlockExplorer);
```
rpanic: We don't need to do this - this is automically taken care of by the Appchain (see ModuleContainer's implementation)

packages/stack/test/graphql/graphql.test.ts
```diff
@@ -154,18 +159,65 @@ describe("graphql client test", () => {
...
+    });
+    await tx.sign();
+    await tx.send();
+
+    const txHash = tx.transaction?.hash().toString()!;
+
+    const preBlockQuery = await appChain.query.explorer.waitTxInclusion(
+      txHash
+    );
+    expect(preBlockQuery.transactionState).toBe(InclusionStatus.PENDING);
```
rpanic: The test here should rather be that the waitTxInclusion returned promise doesn't resolve until we call produceBlock() - only then should it resolve (successfully)

packages/stack/test/graphql/graphql.test.ts
```diff
@@ -154,18 +159,65 @@ describe("graphql client test", () => {
...
+    });
+
+    await tx.sign();
+    await tx.send();
+
+    const block = await trigger.produceBlock();
+    const hash = block?.hash.toString()!;
+    const height = Number(block?.height.toBigInt());
+
+    const blockResult = await appChain.query.explorer.getBlock(hash,height);
```
rpanic: Why do we test with both here? The arguments here are meant as "supply me either hash or height, and I'll find you the corresponding block". Using both here doesn't make much sense. What we should do is call it twice with the hash and then the height and see that the returned block is the correct one.
Because if you look at your implementations, the if statements first check whether `hash !== undefined`, so then the second branch isn't even hit (where the height would be used to fetch).

While we are at it, lets encode this invariant in the type system and make it `{hash: string} | {height: number}`

rpanic (changes requested):

packages/stack/test/graphql/graphql-server.ts
4ytuncs: Removed that, since it is now taken care by appChain.

packages/sequencer/src/helpers/query/BlockExplorer.ts
4ytuncs: Removed the `interval` and `attempts` parameters & hardcoded them.

packages/sdk/src/query/InMemoryBlockExplorer.ts
4ytuncs: Removed returning Pending state.

packages/sdk/src/graphql/GraphqlExplorer.ts
4ytuncs: Done that, promise resolvs or return unknown (which indicates an error)

packages/sdk/src/client/ClientAppChain.ts
4ytuncs: Done: [ac91d26](https://github.com/proto-kit/framework/pull/327/commits/ac91d261f5c9e3dea64a7a0264f7e5dc702d1693)

packages/sdk/src/graphql/GraphqlExplorer.ts
4ytuncs: Done: [da600e2](https://github.com/proto-kit/framework/pull/327/commits/da600e2022e5b512668f71affcf5cc95c633cdec)

packages/sdk/src/graphql/GraphqlExplorer.ts
4ytuncs: Removed GraphqlExplorer.ts, yet it is implemented in the `GraphqlBlockExplorerTransportModule`: [20d1980](https://github.com/proto-kit/framework/pull/327/commits/20d19809e4f24935eb6501a19aebec35ed7ecce6)

packages/stack/test/graphql/graphql.test.ts
4ytuncs: Done: [5c9638d](https://github.com/proto-kit/framework/pull/327/commits/5c9638d67ec0d525316d8082c6cb70cf0b5211fc)

Checks that resolved promise returns the included block.

packages/stack/test/graphql/graphql.test.ts
4ytuncs: Done here: [bbbd10a](https://github.com/proto-kit/framework/pull/327/commits/bbbd10a0eb93e5c5790fd3e4eff43993afcb5243)

Checks that getBlocks function called with height and hash is returning the same block object

packages/sdk/src/graphql/GraphqlExplorer.ts
4ytuncs: I wanted to make it more readable and reusable in the further use. 

Yet it now felt like it can be dealt later, so i'll prefer to remove now here: [77b6799](https://github.com/proto-kit/framework/pull/327/commits/77b67990f4f7eff3072765f3e89361ca5f207982)

packages/sequencer/src/helpers/query/BlockExplorerTransportModule.ts
```diff
@@ -0,0 +1,6 @@
+import { InclusionStatus } from "@proto-kit/api";
+
+export interface BlockExplorerTransportModule {
+  waitTxInclusion(txHash: string): Promise<string>;
```
rpanic: I don't think this name is accurate, since its not "waiting" - that is handled by the BlockExplorerQuery - this one is rather just fetching the transaction's status and returning it right away. I think a naming that reflects that would be less confusing

rpanic (approved):
Only one small thing, otherwise its gtg

packages/sequencer/src/helpers/query/BlockExplorer.ts
4ytuncs: Update: Added them back and added tsdoc for them.

packages/stack/test/graphql/graphql.test.ts
```diff
@@ -168,4 +172,58 @@ describe("graphql client test", () => {
...
+
+    const block = await trigger.produceBlock();
+    const hash = block?.hash.toString()!;
+    const height = Number(block?.height.toBigInt());
+
+    const hashResult = await appChain.query.explorer.getBlock(hash);
+    const heightResult = await appChain.query.explorer.getBlock(height);
+
+    // Blocks should have same transactionsHash.
+    expect(hashResult).toEqual(heightResult);
```
rpanic, line 258: Lets also check that any of them is equal to `block` too though, they could both be undefined and this would pass

packages/sequencer/src/helpers/query/BlockExplorerTransportModule.ts
```diff
@@ -0,0 +1,8 @@
+import { Block } from "../../storage/model/Block";
+
+export type ClientBlock = Pick<Block, "hash" | "previousBlockHash" | "height" | "transactions" | "transactionsHash">;
+
+export interface BlockExplorerTransportModule {
+  waitTxInclusion(txHash: string): Promise<string>;
+  getBlock(param?: string | number): Promise<ClientBlock | undefined>;
```
rpanic: Okay sorry to be picky about this - but there's no way to know what the param is supposed to represent. That its the hash and height people have to guess.
So what I suggested would make this clearer `{hash: string} | {height: number}`. Lmk if you find that dumb

packages/sequencer/src/helpers/query/BlockExplorerQuery.ts
```diff
@@ -0,0 +1,79 @@
...
+      if (status === "INCLUDED") {
+        return { transactionState: InclusionStatus.INCLUDED };
+      }
+
+      remainingAttempts -= 1;
+
+      if (remainingAttempts <= 0) {
+        return { transactionState: InclusionStatus.UNKNOWN };
+      }
+      console.log("baybay");
```
rpanic: :)

packages/sequencer/src/helpers/query/BlockExplorerQuery.ts
```diff
@@ -0,0 +1,79 @@
...
+  public async waitTxInclusion(
+    txHash: string,
+    interval = 1000,
+    maxAttempts = 10
+  ): Promise<{ transactionState: InclusionStatus }> {
+    let remainingAttempts = maxAttempts;
+
+    while (true) {
+      const status = await this.blockExplorer.waitTxInclusion(txHash);
+      console.log("from transport: ", status);
```
rpanic: :)

packages/sequencer/src/helpers/query/BlockExplorerQuery.ts
```diff
@@ -0,0 +1,79 @@
...
+import { BlockExplorerTransportModule, ClientBlock } from "./BlockExplorerTransportModule";
+
+export enum InclusionStatus {
+  UNKNOWN = "unknown",
+  PENDING = "pending",
+  INCLUDED = "included",
+  SETTLED = "settled",
+}
+
+export type TransactionFetcher = (txHash: string) => Promise<InclusionStatus>;
```
rpanic: Afaiu this isn't used anywhere?

packages/sdk/package.json
```diff
@@ -29,7 +29,8 @@
     "@proto-kit/module": "*",
     "@proto-kit/protocol": "*",
     "@proto-kit/sequencer": "*",
-    "o1js": "^1.6.0",
+    "@proto-kit/api": "*",
```
rpanic: We can remove this reference again, now that we don't need it anymore

packages/sequencer/src/helpers/query/BlockExplorerQuery.ts
4ytuncs: :OOOOO 

Im so embarassed.. I used it to test the building stuff, removing asap

packages/sequencer/src/helpers/query/BlockExplorerTransportModule.ts
4ytuncs: Yeah that is what i wanted to as well, it feels like a better develpoer experince , but i didn't feel good visually the way input is given: 

getBlock({hash: '12312321'}) or getBlock({height: 35}) 

So simply it is done like that by aesthetic judgement of mine.

rpanic: I don't mind this at all - I think its pretty common pattern for query functions like these. I with js had named arguments. But generally, I prefer explicitness and safety over dev exp when the tradeoff is potential bugs (which imo with the current typing it is)

packages/sequencer/src/helpers/query/BlockExplorerQuery.ts
4ytuncs: Done..

packages/sequencer/src/helpers/query/BlockExplorerQuery.ts
4ytuncs: Yes - it was intended to be used in some prev implementation.  Removed.

packages/sdk/package.json
4ytuncs: Removed.

packages/stack/test/graphql/graphql.test.ts
4ytuncs: Done here: [8d4e2c5](https://github.com/proto-kit/framework/pull/327/commits/8d4e2c5a14ab6a4b2ff74686947c5744584e728a)

packages/sequencer/src/helpers/query/BlockExplorerTransportModule.ts
4ytuncs: Done: [42caa2e](https://github.com/proto-kit/framework/pull/327/commits/42caa2e65b5be089bce70dcf6f7a423c0c4a3381)

rpanic (approved):

PR #333, opened 2025-11-10
Kadirchan/Cache hashing of transactions

### LINKED ISSUE #322: Cache hashing of transactions
Currently, during block production but also other processes, we hash any given transaction object a bunch of times to determine their id (which is the hash). To improve performance and not do any unnecessary work, we should only hash the transaction once and reuse the resulting hash everywhere else.

kadirchan:
This PR implements cached hash fields into `RuntimeTransaction` and `UnsignedTransaction` classes.

closes #322

### DIFF
--- a/packages/protocol/src/model/transaction/RuntimeTransaction.ts
+++ b/packages/protocol/src/model/transaction/RuntimeTransaction.ts
@@ -13,18 +13,49 @@ export class RuntimeTransaction extends Struct({
   argsHash: Field,
   nonce: UInt64Option,
   sender: PublicKeyOption,
+  hash: Field,
 }) {
+  public static create(input: {
+    methodId: Field;
+    argsHash: Field;
+    nonce: UInt64Option;
+    sender: PublicKeyOption;
+  }) {
+    const { methodId, argsHash, nonce, sender } = input;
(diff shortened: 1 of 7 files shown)


packages/protocol/src/model/transaction/RuntimeTransaction.ts
```diff
@@ -13,18 +13,49 @@ export class RuntimeTransaction extends Struct({
   argsHash: Field,
   nonce: UInt64Option,
   sender: PublicKeyOption,
+  hash: Field,
```
rpanic: I would remove this cashing mechanism, since its used very rarely (only in two circuits) and is provable (in-circuit), so it would have to be constrained right - I think not worth it, so I'd remove this

packages/protocol/src/model/transaction/SignedTransaction.ts
```diff
@@ -26,7 +26,7 @@ export class SignedTransaction extends Struct({
   }
 
   public hash(): Field {
-    return this.transaction.hash();
+    return this.transaction.hash;
```
rpanic: Same, I'd leave this as a function (since its the same as RuntimeTransaction)

PR #335, opened 2025-11-10
Feature/cli graphql docs

stanlou:
### Description

This PR introduces a new **CLI package** to the framework that currently handles **GraphQL documentation generation**.  
Additionally, a **GitHub Actions workflow** has been added to:

- Use the CLI to generate GraphQL docs.  
- Automatically deploy the generated docs to **GitHub Pages**.

### Changes Included

- New **CLI package** under `packages/cli`.  
- **GitHub Actions** workflow for documentation generation and publishing.  
- Basic configuration for GraphQL docs generation.

### Usage

You can use the new CLI both in the **framework repository** and the **starter kit**.

#### In the framework repository
Run the following command to start a fresh AppChain instance and generate GraphQL documentation:
```bash
npx @proto-kit/cli generate-gql-docs -e
```
Optionally, specify the GraphQL server port:
```bash
npx @proto-kit/cli generate-gql-docs -e -p 5050
```
#### In the starter kit
After installing the CLI package, you can generate the GraphQL documentation directly:
```bash
npx @proto-kit/cli generate-gql-docs
```
You can also specify a custom GraphQL endpoint:
```bash
npx @proto-kit/cli generate-gql-docs -u <graphql_url>
```


packages/cli/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@proto-kit/cli",
+  "version": "0.1.0",
+  "type": "module",
+  "main": "./bin/protokit-cli.js",
+  "bin": {
+    "proto-kit": "./bin/protokit-cli.js"
+  },
+  "publishConfig": {
```
rpanic, line 9: Will this cli also be usable from the starter kit? If so - we can leave this public. If not - I'd make it private

packages/cli/src/commands/generateGqlDocs.ts
```diff
@@ -0,0 +1,77 @@
...
+import {
+  GraphqlSequencerModule,
+  GraphqlServer,
+  VanillaGraphqlModules,
+} from "@proto-kit/api";
+import { Runtime } from "@proto-kit/module";
+
+import { generateGqlDocs } from "../utils";
+
+export async function generateGqlDocsCommand(args: {
```
rpanic, line 26: So I wonder - if we want to use this from the starter kit, it'd probably best if we figure out how to make it usable, since we hardcode the appchain instance here. For the starter kit, we'd want this to accept some sort of connection string that'd point to a running appchain having been started in the starter-kit

rpanic: But this can be a follow-up PR as needed

packages/cli/src/commands/generateGqlDocs.ts
```diff
@@ -0,0 +1,77 @@
...
+    });
+
+    await appChain.start();
+    console.log("AppChain started successfully!");
+
+    const gqlUrl = `http://localhost:${port}/graphql`;
+    await generateGqlDocs(gqlUrl);
+  } else {
+    console.log(`Using existing GraphQL endpoint: ${args.url}`);
+    await generateGqlDocs(args.url);
```
rpanic, line 76: We probably shouldn't forget to call `await appChain.close()`, otherwise the process might stay open (though I'm not sure - definitely can't hurt though)

packages/cli/src/commands/generateGqlDocs.ts
rpanic: Ok I'm stupid, you already did that 😂

packages/cli/src/index.ts
```diff
@@ -0,0 +1,47 @@
...
+          default: 8080,
+          describe: "Port for the GraphQL server if creating an AppChain",
+        })
+        .option("url", {
+          alias: "u",
+          type: "string",
+          default: "http://localhost:8080/graphql",
+          describe: "GraphQL endpoint to use if not starting AppChain",
+        })
+        .option("empty", {
```
rpanic, line 28: Isn't "empty" equivalent to the "url" not being set?

.github/workflows/deploy-graphql-docs.yml
```diff
@@ -0,0 +1,69 @@
+name: Generate & Deploy GraphQL Docs
+
+on:
```
rpanic, line 3: Is there a good way to test if this workflow works? Or should we just try it out :)

packages/cli/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "spectaql.config.template.yml"
+  ],
+  "scripts": {
+    "build": "tsc -p tsconfig.json",
+    "lint": "eslint ./src"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "yargs": "17.7.2",
```
rpanic, line 24: Shouldn't spactaql be in the dependency list?

packages/cli/package.json
stanlou: Yes, it can be used from the starter kit.

packages/cli/src/index.ts
stanlou: The empty parameter is used to start the CLI with the hardcoded AppChain, which is mainly intended for cases when running inside the framework.

.github/workflows/deploy-graphql-docs.yml
stanlou: This workflow uses the CLI to generate the GraphQL documentation and then deploys it to GitHub Pages. If the CLI works properly, the workflow should work as expected.

packages/cli/src/commands/generateGqlDocs.ts
stanlou: [Done](https://github.com/proto-kit/framework/blob/f9f5d3b87a663df73b6709ee636f914ebb55c45a/packages/cli/src/commands/generateGqlDocs.ts#L73) .

packages/cli/package.json
stanlou: [Done](https://github.com/proto-kit/framework/blob/f9f5d3b87a663df73b6709ee636f914ebb55c45a/packages/cli/package.json#L26) .

stanlou: [Done](https://github.com/proto-kit/framework/blob/f9f5d3b87a663df73b6709ee636f914ebb55c45a/packages/cli/package.json#L26) .

rpanic (approved):

rpanic:
@stanlou Lets fix the CI, then we're gtg

PR #339, opened 2025-11-11
Add sequencer id to job payloads

### LINKED ISSUE #314: Add unique sequencer id to job payload
Currently, after a sequencer has been forcefully shutdown. There can be two cases that are connected with the bullmq retaining their state and not being restarted:
1. Sequencer restarts and receives results from tasks dispatched by the previous instance
2. Worker starts and sees old tasks and starts computing them

(1) should be handled by attaching a instance-unique id to the payload and filtering through that for every received response.
This also leads to (2) not being a big issue anymore as those results just get thrown away later.

Before implementing this, research if it is actually a problem

Raunaque97:
This PR adds a unique `sequencer.id` to a sequencer instance and includes it in all task payloads. 
So the Flow class can now filters out responses from old sequencer instances.

Fixes #314


.vscode/launch.json
rpanic, line 1: This is empty - so we should probably remove it?

packages/sequencer/src/worker/flow/Flow.ts
```diff
@@ -177,10 +185,11 @@ export class Flow<State> implements Closeable {
 @injectable()
 export class FlowCreator {
   public constructor(
-    @inject("TaskQueue") private readonly queueImpl: TaskQueue
+    @inject("TaskQueue") private readonly queueImpl: TaskQueue,
+    @inject("Sequencer") private readonly sequencer: Sequencer<any>
```
rpanic: Okay so I think there is a pattern how we can solve this better.
What I would do, is I would create a `class SequencerIdProvider` that has `getSequencerId()`.
Then, we annotate that class using `@injectable() and @scoped(Lifecycle.Singleton)`.
This way, we only ever create one instance of that class (this what Lifecycle.Singleton does) and we can inject it very easily via `private readonly sequencerIdProvider: SequencerIdProvider` - note, without `@inject("")`, since we want to inject the class directly - tsyringe will create the instance for us

packages/sequencer/test/protocol/production/flow/ReductionTaskFlow.test.ts
```diff
@@ -90,6 +90,7 @@ describe("ReductionTaskFlow", () => {
     di.register("TaskQueue", {
       useValue: queue,
     });
+    di.register("Sequencer", { useValue: { id: "1" } });
```
rpanic: Using the pattern I described earlier - we get rid of this workaround as well

packages/sequencer/src/worker/flow/Flow.ts
Raunaque97: moved id to sequencerIdProvider

packages/sequencer/src/sequencer/executor/Sequencer.ts
```diff
@@ -81,6 +86,12 @@ export class Sequencer<Modules extends SequencerModulesRecord>
 
     this.useDependencyFactory(MethodIdFactory);
 
+    // Drain all task queues to clear stale tasks from previous sequencer instances
+    if (this.container.isRegistered("TaskQueue")) {
+      const taskQueue = this.container.resolve<TaskQueue>("TaskQueue");
```
rpanic: I thought I added a comment here but apparently didn't (please say if we already talked about it).
But I'd much rather see this logic here implemented in the start() function that the TaskQueue implements (SequencerModule requires that anyways, and it is automatically called by the sequencer during startup)

rpanic (approved):
Approved given fix of the one new comment

Issue #302, opened 2025-04-03
Integrate Signer module into Settlement


rpanic:
Currently, the settlement implementation requires a PrivateKey to be passed in to sign transactions. Replace this mechanism by letting the user supply a Signer module that the settlement process uses. 
This enables things like remote signing and is generally a better security practice since it makes it harder to accidentally leak a private key

4ytuncs:
[Feat/signer integration #354](https://github.com/proto-kit/framework/pull/354) closes this issue.

Issue #301, opened 2025-04-03
Introduce separate signer for sequencer


rpanic:
Make the Signer used by the sequencer for things like Settlement different from the client-used Signer to distinguish client from sequencer signer instances

4ytuncs:
https://github.com/proto-kit/framework/pull/354 closes this issue.

PR #354, opened 2025-11-21
Feat/signer integration

4ytuncs:
## Signer module integration to Settlement module

This PR closes issues #302 and #301. 

### Changes

* `MinaSigner` interface and `InMemoryMinaSigner` module is implemented. This module is used as a key managament system, where private keys are held here and signings are done from here. This enables settlement transactions to be signed inside `InMemoryMinaSigner`, without exposing private keys to any other part of the memory. 
*  Besides contract keys, any Token Bridge key and their owner/admin keys are kept in the signer and retrieved by using a `Map`. This enables users to pass the regarding public key only - signer will find the private key and execute signings. 
* Settlement contract used to take settlementKey as input and assert that public key matches with the relevant contract. To keep private key safe, now a signature created by settlement contract key.

### Commit Details 
   * Implemented Signer interface and `InMemoryMinaSigner` : da8204c
   * Updated SettlementUtils.ts: b58840d 
   * Updated bridging module: c72ce51
   * Updated `settlementTestFn`: 91b42b7
   * Added signature creation with a specified key on signer config e24c1e2

### DIFF
--- a/packages/protocol/src/settlement/contracts/SettlementSmartContract.ts
+++ b/packages/protocol/src/settlement/contracts/SettlementSmartContract.ts
@@ -69,14 +69,7 @@ export interface SettlementContractType {
     args: DeployArgs | undefined,
     permissions: Permissions,
     sequencer: PublicKey,
-<<<<<<< HEAD
     dispatchContract: PublicKey
-=======
-    dispatchContract: PublicKey,
-    bridgeContract: PublicKey,
-    contractKey: PublicKey,
-    contractSignature: Signature
->>>>>>> de1c1c20 (refactor: update address check in settlement contract)
   ) => Promise<void>;
 
   assertStateRoot: (root: Field) => AccountUpdate;
@@ -253,39 +246,13 @@ export abstract class SettlementSmartContractBase extends TokenContract {
 
   protected async initializeBase(
     sequencer: PublicKey,
-<<<<<<< HEAD
     dispatchContract: PublicKey
-=======
-    dispatchContract: PublicKey,
-    bridgeContract: PublicKey,
-    contractKey: PublicKey,
-    contractSignature: Signature,
->>>>>>> de1c1c20 (refactor: update address check in settlement contract)
   ) {
     this.sequencerKey.set(sequencer.x);
     this.stateRoot.set(LinkedMerkleTree.EMPTY_ROOT);
     this.blockHashRoot.set(Field(BlockHashMerkleTree.EMPTY_ROOT));
     this.networkStateHash.set(NetworkState.empty().hash());
     this.dispatchContractAddressX.set(dispatchContract.x);
-<<<<<<< HEAD
-=======
-
-    const { DispatchContract } = SettlementSmartContractBase.args;
-    const contractInstance = new DispatchContract(dispatchContract);
-    await contractInstance.initialize(this.address);
-
-    // Deploy bridge contract for $Mina
-    await this.deployTokenBridge(
-      this.tokenId,
-      bridgeContract,
-      dispatchContract,
-      true
-    );
-
-    contractSignature.verify(contractKey, contractKey.toFields());
-    // contractKey.toPublicKey().assertEquals(this.address);
-
->>>>>>> de1c1c20 (refactor: update address check in settlement contract)
   }
 
   protected async settleBase(
@@ -474,26 +441,8 @@ export class SettlementSmartContract
     await this.initializeBase(sequencer, dispatchContract);
   }
 
-<<<<<<< HEAD
   @method async approveBase(forest: AccountUpdateForest) {
     this.checkZeroBalanceChange(forest);
-=======
-  @method
-  public async initialize(
-    sequencer: PublicKey,
-    dispatchContract: PublicKey,
-    bridgeContract: PublicKey,
-    contractKey: PublicKey,
-    contractSignature: Signature
-  ) {
-    await this.initializeBase(
-      sequencer,
-      dispatchContract,
-      bridgeContract,
-      contractKey,
-      contractSignature
-    );
->>>>>>> de1c1c20 (refactor: update address check in settlement contract)
   }
 
   @method
--- a/packages/protocol/src/settlement/contracts/SettlementSmartContract.ts
+++ b/packages/protocol/src/settlement/contracts/SettlementSmartContract.ts
@@ -69,7 +69,14 @@ export interface SettlementContractType {
     args: DeployArgs | undefined,
     permissions: Permissions,
     sequencer: PublicKey,
+<<<<<<< HEAD
     dispatchContract: PublicKey
+=======
+    dispatchContract: PublicKey,
+    bridgeContract: PublicKey,
+    contractKey: PublicKey,
+    contractSignature: Signature
+>>>>>>> de1c1c20 (refactor: update address check in settlement contract)
   ) => Promise<void>;
 
   assertStateRoot: (root: Field) => AccountUpdate;
@@ -246,13 +253,39 @@ export abstract class SettlementSmartContractBase extends TokenContract {
 
   protected async initializeBase(
     sequencer: PublicKey,
+<<<<<<< HEAD
     dispatchContract: PublicKey
+=======
+    dispatchContract: PublicKey,
+    bridgeContract: PublicKey,
+    contractKey: PublicKey,
+    contractSignature: Signature,
+>>>>>>> de1c1c20 (refactor: update address check in settlement contract)
   ) {
     this.sequencerKey.set(sequencer.x);
     this.stateRoot.set(LinkedMerkleTree.EMPTY_ROOT);
     this.blockHashRoot.set(Field(BlockHashMerkleTree.EMPTY_ROOT));
     this.networkStateHash.set(NetworkState.empty().hash());
     this.dispatchContractAddressX.set(dispatchContract.x);
+<<<<<<< HEAD
+=======
+
+    const { DispatchContract } = SettlementSmartContractBase.args;
+    const contractInstance = new DispatchContract(dispatchContract);
+    await contractInstance.initialize(this.address);
+
+    // Deploy bridge contract for $Mina
+    await this.deployTokenBridge(
+      this.tokenId,
+      bridgeContract,
+      dispatchContract,
+      true
+    );
+
+    contractSignature.verify(contractKey, contractKey.toFields());
+    // contractKey.toPublicKey().assertEquals(this.address);
+
+>>>>>>> de1c1c20 (refactor: update address check in settlement contract)
   }
 
   protected async settleBase(
@@ -441,8 +474,26 @@ export class SettlementSmartContract
     await this.initializeBase(sequencer, dispatchContract);
   }
 
+<<<<<<< HEAD
   @method async approveBase(forest: AccountUpdateForest) {
     this.checkZeroBalanceChange(forest);
+=======
+  @method
+  public async initialize(
+    sequencer: PublicKey,
+    dispatchContract: PublicKey,
+    bridgeContract: PublicKey,
+    contractKey: PublicKey,
+    contractSignature: Signature
+  ) {
+    await this.initializeBase(
+      sequencer,
+      dispatchContract,
+      bridgeContract,
+      contractKey,
+      contractSignature
+    );
+>>>>>>> de1c1c20 (refactor: update address check in settlement contract)
   }
 
   @method
--- a/packages/sequencer/src/sequencer/SequencerStartupModule.ts
+++ b/packages/sequencer/src/sequencer/SequencerStartupModule.ts
@@ -22,7 +22,6 @@ import {
 import { VerificationKeyService } from "../protocol/runtime/RuntimeVerificationKeyService";
 import type { MinaBaseLayer } from "../protocol/baselayer/MinaBaseLayer";
 import { NoopBaseLayer } from "../protocol/baselayer/NoopBaseLayer";
-import { MinaSigner } from "../settlement/MinaSigner";
 
 import { SequencerModule, sequencerModule } from "./builder/SequencerModule";
 import { Closeable, closeable } from "./builder/Closeable";
@@ -45,8 +44,6 @@ export class SequencerStartupModule
     private readonly baseLayer: MinaBaseLayer | undefined,
     @inject("AreProofsEnabled")
     private readonly areProofsEnabled: AreProofsEnabled,
-    @inject("SettlementSigner", { isOptional: true })
-    private readonly signer: MinaSigner | undefined
   ) {
     super();
   }
--- a/packages/sequencer/src/sequencer/SequencerStartupModule.ts
+++ b/packages/sequencer/src/sequencer/SequencerStartupModule.ts
@@ -21,7 +21,6 @@ import {
 } from "../protocol/production/tasks/CircuitCompilerTask";
 import { VerificationKeyService } from "../protocol/runtime/RuntimeVerificationKeyService";
 import type { MinaBaseLayer } from "../protocol/baselayer/MinaBaseLayer";
-import { SettlementUtils } from "../settlement/utils/SettlementUtils";
 import { NoopBaseLayer } from "../protocol/baselayer/NoopBaseLayer";
 import { MinaSigner } from "../settlement/MinaSigner";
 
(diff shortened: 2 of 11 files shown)


packages/protocol/src/settlement/contracts/SettlementSmartContract.ts
```diff
@@ -273,7 +274,8 @@ export abstract class SettlementSmartContractBase extends TokenContract {
       true
     );
 
-    contractKey.toPublicKey().assertEquals(this.address);
+    contractSignature.verify(contractKey, contractKey.toFields());
```
rpanic: This i pretty much a no-op because contractKey could be anything :)
Change to `contractSignature.verify(this.address, contractKey.toFields());`

packages/sequencer/src/settlement/MinaSigner.ts
```diff
@@ -0,0 +1,130 @@
...
+
+import { SequencerModule } from "../sequencer/builder/SequencerModule";
+
+export interface SignTxOptions {
+  pubKeys?: PublicKey[];
+  additionalKeys?: PrivateKey[];
+  signWithContract?: boolean;
+}
+export interface MinaSigner {
+  getSignerAddress(): PublicKey;
```
rpanic: Can we make this name a bit more descriptive? We have a lot of signers here - which one is this referring to?

packages/sequencer/src/settlement/MinaSigner.ts
```diff
@@ -0,0 +1,130 @@
+import { Field, PrivateKey, PublicKey, Signature, Transaction } from "o1js";
+import { noop } from "@proto-kit/common";
+import { injectable } from "tsyringe";
+
+import { SequencerModule } from "../sequencer/builder/SequencerModule";
+
+export interface SignTxOptions {
+  pubKeys?: PublicKey[];
+  additionalKeys?: PrivateKey[];
+  signWithContract?: boolean;
```
rpanic: I assume this flag should say whether we are in signed or proven settlement?
If so, we already have SettlementUtils.isSignedSettlement() which we should use here

packages/sequencer/src/settlement/BridgingModule.ts
```diff
@@ -398,11 +400,10 @@ export class BridgingModule {
         }
       );
 
-      const signedTx = this.utils.signTransaction(
-        tx,
-        [feepayer],
-        options.contractKeys
-      );
+      const signatureOnes = options.contractKeys.map((x) => x.toPublicKey());
```
rpanic: What is the thought here, can we drop the options' privatekeys and instead use the ones from the signer?

packages/sequencer/src/settlement/utils/SettlementUtils.ts
```diff
@@ -4,20 +4,31 @@ import {
...
   UInt32,
 } from "o1js";
 import { AreProofsEnabled, mapSequential } from "@proto-kit/common";
 
 import type { MinaBaseLayer } from "../../protocol/baselayer/MinaBaseLayer";
+import { MinaSigner } from "../MinaSigner";
+
+interface SignTransactionOptions {
+  signingWithSignatureCheck?: PublicKey[];
+  signingPublicKeys?: PublicKey[];
```
rpanic, line 17: I am also confused here - what do these 4 different arrays do exactly different from each other? I assume this can be reduced somewhat

packages/sequencer/src/settlement/utils/SettlementUtils.ts
```diff
@@ -31,29 +42,60 @@ export class SettlementUtils {
...
-   * Sign transaction with two variants:
-   *
-   * - If it normal settlement (proofs enabled or mock proofs):
-   *   Sign the transaction normally
-   *
-   * - If it is signed settlement:
-   *   Signed the transactions and make all contract AUs where a private key is known
-   *   via the contractKeys param require a signature and sign them using that key
-   */
+  public getSigner(): PublicKey {
```
rpanic: Same as the comment above - I think this gives you the fee payer pubkey? If so, lets call the function that way as well

packages/sequencer/src/settlement/utils/SettlementUtils.ts
```diff
@@ -31,29 +42,60 @@ export class SettlementUtils {
...
+    return this.signer.getSignerAddress();
+  }
+
+  public sign(signatureData: Field[]): Signature {
+    return this.signer.sign(signatureData);
+  }
+
+  public signMessageWithContract(signatureData: Field[]): Signature {
```
rpanic: Same here - I know what it's for because we designed the mechanism with the signature for initialize, but for another reader, this is confusing as hell.
Also - I think this should be general and sign given a argument that's the pubkey of the contract

packages/sequencer/src/settlement/utils/SettlementUtils.ts
rpanic: And yes, I realize that most of this is old baggage from my original implementation, but we should make it clearer for this refactor :)) Sorry about my past self

packages/sequencer/src/settlement/utils/SettlementUtils.ts
```diff
@@ -31,29 +42,60 @@ export class SettlementUtils {
...
-    this.requireSignatureIfNecessary(
-      tx,
-      contractKeyArray.map((key) => key.toPublicKey()),
-      preventNoncePreconditionFor
-    );
-    return tx.sign([...pks, ...contractKeyArray]);
+    options: SignTransactionOptions = {}
+  ) {
+    const {
+      signingWithSignatureCheck = [],
```
rpanic, line 35: I think, now that we have the Signer separately, we can ask it for most of the information that we give it here with the options.
Like - the signer should know which keys are EOAs and which are contracts - and in that case should treat all contract keys with respect to isSignedSettlement. Of course we also must consider the cases for deployment and initialization

rpanic (changes requested):

packages/sequencer/src/sequencer/SequencerStartupModule.ts
```diff
@@ -142,10 +145,13 @@ export class SequencerStartupModule
 
     // TODO Find a way to generalize this or at least make it nicer - too much logic here
     const isSignedSettlement =
-      this.baseLayer !== undefined && !(this.baseLayer instanceof NoopBaseLayer)
+      this.baseLayer !== undefined &&
```
rpanic: I'm not very happy with this part of the code. There could exist an edge case where the user forgets to configure the SettlementModule and then this silently resolves to undefined.
This is unrelated to your PR though - should we move isSignedSettlement() out from utils and into MinaBaseLayer?

packages/sequencer/src/settlement/MinaSigner.ts
4ytuncs: im not sure about it, what i was trying to choose using contract keys as they differed here:

https://github.com/proto-kit/framework/blob/088c3e979f4294b50427fa3a065a44acdc804058/packages/sequencer/test/settlement/Settlement.ts#L407

https://github.com/proto-kit/framework/blob/088c3e979f4294b50427fa3a065a44acdc804058/packages/sequencer/test/settlement/Settlement.ts#L363

Also i realise i chose a bad naming, `signWithContracts` here choose whether if  `settlement`, `dispatch` and `minaBridge`  contracts are used for signing or not. As much as i see isSignedSettlement do not directly used to include/exclude settlement,dispatch& bridge contracts.

In one of our chats you said:

"Ah yeah last thing - what the utils.signTransaction() now does is that it receives two arrays of keys, one that it always needs, and one that it only needs for signed settlements (those are the contract keys) and then it chooses based on the settlement type"

Yet the use of contract keys differ by choosing `settlementModule.signTransaction` or `settlementModule.utils.signTransaction`, where .utils one do not force us to use any key necessarily.

packages/sequencer/src/settlement/BridgingModule.ts
4ytuncs: This was the change i mentioned to you that i'd change. I wanted to fix this together with the tasks review that'll come with review.

packages/sequencer/src/settlement/utils/SettlementUtils.ts
4ytuncs: I had to add tsdoc-like comment lines, will be adding in this PR. 

signingWithSignatureCheck: Those are the public keys which will be included to the `requireIfSignatureNecessary` function.
signingPublicKeys and additionalKeys: those are the ones that are in `pks` previously. Actually there should be only one key array in `PublicKey`, yet in this [Settlement tests](https://github.com/proto-kit/framework/blob/088c3e979f4294b50427fa3a065a44acdc804058/packages/sequencer/test/settlement/Settlement.ts#L499) `userKey` is used as a simulation of end user - so actually `additionalKey` is a parameter that'll just used in tests, should be removed 👍 

Update: Actually we must either keep `additionalKeys` or register it as a key in our `InMemoryMinaSigner`. I suggest the second one - which is beneficial for future. any key that users would want to add in the runtime (like registering a token contract!) can add it.

packages/sequencer/src/settlement/utils/SettlementUtils.ts
4ytuncs: Yes, actually i need your opinion here: 

I didn't want to add contractKeys to the keyMap at the beginning, but it is beneficial for cases where such singular/selective signing operations might be needed. I will update this function in that manner, so it will provide signatures with the given public keys - and it will not be restricted to the contract keys only for better development experience, wdyt?

packages/sequencer/src/sequencer/SequencerStartupModule.ts
4ytuncs: I agree. It can be directly used inside `MinaBaseLayer`, so it is unnecessary to import/inject `MinaBaseLayer` and `areProofsEnabled` to other modules just to initialize `SettlementUtils`.

packages/sequencer/src/settlement/MinaSigner.ts
4ytuncs: Also, previously what we had was two proxy functions for signing: one in `SettlementUtils` and the other in `SettlementModule`. The one in the `SettlementModule` make it necessary to use contract keys in `requireSignatureIfNecessary`, depending on the result of `isSignedSettlement`. 

Instead of implementing two functions in two separate classes, i've addded one flag to do the same thing actually.

packages/sequencer/src/settlement/utils/SettlementUtils.ts
rpanic: I think this is the right approach tbh, as we've discussed on slack too, we should make this as dynamic and flexible as possible. So we should go for your suggestion

rpanic: I think this is the right approach tbh, as we've discussed on slack too, we should make this as dynamic and flexible as possible. So we should go for your suggestion

packages/protocol/src/settlement/contracts/SettlementSmartContract.ts
4ytuncs: Fixed here: ed2dbda

packages/sequencer/src/settlement/MinaSigner.ts
4ytuncs: I think using the name 'feepayer' again would be fine for this one. 
Done here: 9572c3e

packages/sequencer/src/settlement/MinaSigner.ts
4ytuncs: Updated the mechanism as discussed: `isSignedSettlement` is used for our purposes and signWithContract is removed: 24f7057

packages/sequencer/src/settlement/BridgingModule.ts
4ytuncs: Done here: b3cb936

packages/sequencer/src/settlement/utils/SettlementUtils.ts
4ytuncs: `additionalKeys` is removed, now keys are registered: 84a1bab

packages/sequencer/src/settlement/utils/SettlementUtils.ts
4ytuncs: Done here: 9572c3e

packages/sequencer/src/settlement/utils/SettlementUtils.ts
4ytuncs: Done here, now we can create signatures with any key exists in the InMemoryMinaSigner: 7570cd5

packages/sequencer/src/sequencer/SequencerStartupModule.ts
4ytuncs: Done here: 92a8814

packages/sequencer/src/settlement/MinaSigner.ts
```diff
@@ -0,0 +1,141 @@
...
+  pubKeys?: PublicKey[];
+}
+export interface MinaSigner {
+  getFeepayerKey(): PublicKey;
+
+  getContractAddresses(): PublicKey[];
+
+  sign(signatureData: Field[]): Signature;
+
+  signWithKey(publicKey: string, signatureData: Field[]): Signature;
```
rpanic: Since we usually treat pubkeys using the PublicKey type, I understand why you used string here, but imo this leaks from the implementation - we should call `.toBase58()` internally and accept a `PublicKey` here

packages/sequencer/src/settlement/MinaSigner.ts
```diff
@@ -0,0 +1,141 @@
...
+      }
+      privateKeys.push(privKey);
+    }
+
+    const keys = [this.config.feepayer, ...privateKeys];
+    return tx.sign(keys);
+  }
+
+  public async start() {
+    this.initializeKeyMap();
```
rpanic, line 332: We can actually do this in the constructor as well iirc which would be a better place to initialize this

packages/sequencer/src/settlement/MinaSigner.ts
```diff
@@ -0,0 +1,141 @@
...
+    options?: SignTxOptions
+  ): Transaction<false, true>;
+
+  registerKey(privateKey: PrivateKey): PublicKey;
+}
+
+export interface InMemorySignerConfig {
+  feepayer: PrivateKey;
+  contractKeys: PrivateKey[];
+  tokenControllers?: PrivateKey[];
```
rpanic: I think we will be fine by having only feepayer, contractKeys and additionalKeys. Because we don't seem to make any implementation-wise distinction between the two latter ones

packages/sequencer/src/settlement/utils/SettlementUtils.ts
```diff
@@ -4,64 +4,81 @@ import {
...
-  public isSignedSettlement(): boolean {
-    return (
-      !this.areProofsEnabled.areProofsEnabled &&
-      !this.baseLayer.isLocalBlockChain()
-    );
+  public getFeepayerKey(): PublicKey {
+    return this.signer.getFeepayerKey();
+  }
+
+  public sign(signatureData: Field[]): Signature {
```
rpanic: We probably shouldn't create those function and in the `SettlementModule` inject and access the signer directly (except the signTransaction of course)

packages/sequencer/src/settlement/SettlementModule.ts
```diff
@@ -300,23 +255,21 @@ export class SettlementModule
               verificationsKeys.SettlementSmartContract.verificationKey,
           },
           permissions.settlementContract(),
-          feepayerKey.toPublicKey(),
-          dispatchKey.toPublicKey()
+          feepayer,
```
rpanic, line 252: You can actually remove the feepayer here (if we're rebased on the latest develop). With the restructuring I did, this isn't necessary anymore.
See https://github.com/proto-kit/framework/pull/366/files#diff-d024c7b9b45b21b301336a6de5a0e8ca50e41b026855affcb8acea651b16cbd0R72

packages/sequencer/src/settlement/MinaSigner.ts
rpanic: For context - the account to "which publickey corresponds to which bridge/token" is done by the SettlementModule directly, but only with pubkeys

rpanic (changes requested):

packages/sequencer/src/settlement/MinaSigner.ts
4ytuncs: Done here: [e24c1e2](https://github.com/proto-kit/framework/pull/354/commits/e24c1e24a9d1787c06b91407bc6b858db884b457)

packages/sequencer/src/settlement/utils/SettlementUtils.ts
4ytuncs: Done here, removed unnecessary functions as much as possible : [b440748](https://github.com/proto-kit/framework/pull/354/commits/b440748f55a66712c21e97664a3a7ce5e166bdfd)

packages/sequencer/src/settlement/SettlementModule.ts
4ytuncs: How? feepayer is simply sequencer: https://github.com/proto-kit/framework/blob/feat/signer-integration/packages/sequencer/test/settlement/Settlement.ts#L176 

which is needed in your update

packages/sequencer/src/settlement/MinaSigner.ts
4ytuncs: Done here: [a81983b](https://github.com/proto-kit/framework/pull/354/commits/a81983b70efff63cec9d981af9bdea3df9aae17b)

packages/sequencer/src/settlement/MinaSigner.ts
4ytuncs: When i move it to the constructor, i get this error:

```
Cannot inject the dependency "messageService" at position #1 of "BlockProducerModule" constructor. Reason:
        Cannot inject the dependency "settlementModule" at position #4 of "IncomingMessagesService" constructor. Reason:
            Cannot inject the dependency "signer" at position #6 of "SettlementModule" constructor. Reason:
```

Since config is not initialized, this do not work in constructor execution step. It works in later steps when start() is executed.

rpanic (approved):

PR #360, opened 2025-11-25
Added retry logic to minaTransactionSender

### LINKED ISSUE #340: Handling of non-inclusion of L1 transactions due to low fee
The settlement process sends a variety of transactions to the Mina L1. This includes settlement transactions, but also various transactions that have to do with bridging. Most of that functionality is currently handled by the MinaTransactionSender and called by the SettlementModule and BridgingModule. 
However, the L1 might not include sent transactions, in most cases because the fee is too low and/or the network is congested. 
When we detect this to be the case, we want to re-submit the same transaction, but with a higher fee.

Raunaque97:
closes #340 

- refactored minaTransactionSender to use a persistent Storage.
- added a simple retry Strategy to increase fees by a multiple and retry every few minutes if the transaction is not included.
- also handles the case if the sequencer restarts and the last sent transaction statuses were not updated in DB.


packages/persistance/prisma/schema.prisma
```diff
@@ -148,3 +148,15 @@ model IncomingMessageBatch {
 
   messages IncomingMessageBatchTransaction[]
 }
+
+model PendingL1Transaction {
```
rpanic, line 152: You forgot to create a migration for this new schema change - do this by cd-ing into the persistence pkg, then run `prisma migrate dev` and it'll you for a name and migrate

rpanic:
So what I just discovered: When we setFee(), o1js erases all previous signatures in order for the tx to be signed again. And this makes sense - for the feepayer and _some_ accountupdates, o1js uses the so-called "full commitment". This is simply just the hash of the whole transaction, including all accountupdates and the feepayer, whereas the non-full commitment would just the be hash of the current accountupdate. 
Now, accountupdates can actually specify whether they want to use the full commitment or not (this is a field in the accountupdate struct that mina will receive as part of the transaction). And whatever commitment you choose, you signature or proof will only be valid given that exact commitment. it is kinda the basis on which you sign or prove. Now why is this important?
Signing or proving over a full commitment prevent replay attacks, like in other chains. But mina is a bit special, and unlike other chains, we sign a bunch of sub-parts individually. And so generally, in order to prevent somebody to rip out a part of our tx from the L1 mempool and replay it in their own specially-crafted malicious transaction and frontrun ours, we sign over the full commitment. 
For resending transactions with a higher fee, this is a bit annoying - we don't want to re-prove the entire tx just because we changed the fee, right?
As I've mentioned above, luckily we can choose to opt-out of using the full commitment and use only the current accountupdate's commitment instead. But this is only possible in some distinct scenarios. My initial suspicious is that this is the case iff the accountupdate's validity is directly dependent on exclusively other full-commitment based accountupdates within the same transaction. But I'll have to think more on that definition.
Anyways, what we have to do before this PR will work without re-proving some accountupdates, is to use no full commitment generally. And for us to be able to do this, we need to examine the protocol and see if we can do this safely

packages/sequencer/src/settlement/transactions/DefaultL1TransactionRetryStrategy.ts
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  public async prepareRetryTransaction(
+    record: PendingL1TransactionRecord
+  ): Promise<Transaction<any, false>> {
+    const tx = Transaction.fromJSON(JSON.parse(record.transactionJson));
+    const currentFee = tx.transaction.feePayer.body.fee;
+    const newFee = UInt64.from(this.bumpFee(Number(currentFee.toBigInt())));
+    tx.setFee(newFee);
+    // Delay if needed
```
rpanic: Can you explain what this delay here accomplishes? I'd expect this function to only craft the new transaction and return it...

packages/sequencer/src/storage/repositories/PendingL1TransactionStorage.ts
```diff
@@ -0,0 +1,34 @@
...
+  | "sent"
+  | "included"
+  | "failed";
+
+export interface PendingL1TransactionRecord {
+  sender: string;
+  nonce: number;
+  attempts: number;
+  status: PendingL1TransactionStatus;
+  transactionJson: string;
```
rpanic: Couldn't this struct here hold the transaction directly? This way we don't have to parse it at all for in-memory mode, and then the Mapper can handle parsing for the Prisma variant

packages/persistance/prisma/schema.prisma
```diff
@@ -148,3 +148,15 @@ model IncomingMessageBatch {
...
+  sender         String
+  nonce          Int
+  attempts       Int
+  status         String   @db.VarChar(32)
+  transaction    Json     @db.Json
+  lastError      String?
+  sentAt         DateTime? 
+
+  @@id([sender, nonce])
```
rpanic: I'm not against this schema, but I think for the reorg support we'll have to change this - but that will happen when we tackle that piece of work.

rpanic: What I'd love to see would be if we could add foreign-key out of settlement here. So in the Settlement entity, add a reference to the PendingTransaction so that we can establish that link

packages/persistance/src/services/prisma/PrismaPendingL1TransactionStorage.ts
```diff
@@ -0,0 +1,103 @@
...
+    const { prismaClient } = this.connection;
+    await prismaClient.pendingL1Transaction.update({
+      where: {
+        sender_nonce: {
+          sender,
+          nonce,
+        },
+      },
+      data: {
+        ...(updates.attempts !== undefined && { attempts: updates.attempts }),
```
rpanic, line 41: Why do you have this here? Since it's partial already, couldn't you just use the `updates` object here directly?

package.json
```diff
@@ -74,6 +74,7 @@
     ]
   },
   "dependencies": {
+    "clean": "^4.0.2",
```
rpanic: This dependency isn't used afaict

packages/sequencer/src/settlement/transactions/DefaultL1TransactionRetryStrategy.ts
```diff
@@ -0,0 +1,90 @@
...
+    return true;
+  }
+
+  public async prepareRetryTransaction(
+    record: PendingL1TransactionRecord
+  ): Promise<Transaction<any, false>> {
+    const tx = Transaction.fromJSON(JSON.parse(record.transactionJson));
+    const currentFee = tx.transaction.feePayer.body.fee;
+    const newFee = UInt64.from(this.bumpFee(Number(currentFee.toBigInt())));
+    tx.setFee(newFee);
```
rpanic: As mentioned in the comment earlier, we need to figure something out to prevent re-proving

packages/sequencer/src/settlement/transactions/DefaultL1TransactionRetryStrategy.ts
```diff
@@ -0,0 +1,90 @@
...
+
+const DEFAULT_RETRY_CONFIG: Required<TransactionRetryConfig> = {
+  maxAttempts: 3,
+  feeMultiplier: 1.1,
+  maxFee: 10 * 1e9,
+  retryDelayMs: 60 * 1000, // 1 minute
+};
+
+@sequencerModule()
+export class DefaultL1TransactionRetryStrategy
```
rpanic, line 29: I like this encapsulation!

packages/sequencer/src/settlement/transactions/MinaTransactionSender.ts
```diff
@@ -31,114 +33,69 @@ export type TxSendResult<Input extends "sent" | "included" | "none"> =
...
-    @inject("BaseLayer") private readonly baseLayer: MinaBaseLayer
-  ) {}
+    @inject("BaseLayer") private readonly baseLayer: MinaBaseLayer,
+    @inject("PendingL1TransactionStorage")
+    private readonly pendingStorage: PendingL1TransactionStorage,
+    @inject("L1TransactionRetryStrategy")
+    private readonly retryStrategy: L1TransactionRetryStrategy,
+    @inject("MinaSigner") private readonly signer: MinaSigner
+  ) {
+    void this.startPolling();
```
rpanic: Lets not void this completely, instead of a while(true), lets create a interval, save the id, and then delete the interval in `close()` (which we have to add using `@closeable()` and `implements Closeable` on this class).
If we don't do this, appchain.close() won't end the process bcs there's still this dangling promise running in the background - this is especially annoying since tests won't terminate

rpanic (changes requested):
This is great! I have a few minor requests and questions, but I like it generally

packages/sequencer/src/settlement/transactions/DefaultL1TransactionRetryStrategy.ts
Raunaque97: This is used to add a delay between retries, like retrying a transaction every 5 minutes.

packages/sequencer/src/settlement/transactions/DefaultL1TransactionRetryStrategy.ts
Raunaque97: 😀

packages/sequencer/src/protocol/baselayer/MinaBaseLayer.ts
```diff
@@ -78,6 +94,17 @@ export class MinaBaseLayer
         useClass: MinaTransactionSender,
       },
 
+      TxStatusWaiter: {
+        useClass: TxStatusWaiter,
+      },
+
+      L1TransactionDispatcherConfig: {
```
rpanic: So - the pattern that we should use here is by using `useFactory` instead of `useValue` and then returning the Dispatcher instance - so inside the factory, resolve the dispatcher from the given dependencycontainer, then setting the config

packages/sequencer/src/settlement/transactions/MinaTransactionSender.ts
```diff
@@ -1,146 +1,125 @@
...
     private readonly simulator: MinaTransactionSimulator,
-    @inject("BaseLayer") private readonly baseLayer: MinaBaseLayer
-  ) {}
+    @inject("BaseLayer") private readonly baseLayer: MinaBaseLayer,
+    @inject("PendingL1TransactionStorage")
+    private readonly pendingStorage: PendingL1TransactionStorage,
+    @inject("SettlementSigner") private readonly signer: MinaSigner,
+    @inject("FeeStrategy") private readonly feeStrategy: FeeStrategy,
+    private readonly dispatcher: L1TransactionDispatcher,
+    @inject("TxStatusWaiter") private readonly waiter: TxStatusWaiter
```
rpanic: If I see it correctly, the waiter is only used inside the MinaTransactionSender and is not supposed to be overrideable - therefore I'd suggest resolving by class reference, not by token here

packages/sequencer/src/settlement/transactions/MinaTransactionSender.ts
```diff
@@ -1,146 +1,125 @@
...
+  ) {
+    this.dispatcher.start();
```
rpanic, line 45: We can use @startable() here (like closeable) and then start the dispatcher in there

packages/sequencer/src/settlement/transactions/L1TransactionDispatcher.ts
```diff
@@ -0,0 +1,279 @@
...
+          await this.tick();
+        });
+      } catch (e) {
+        log.error("Error in L1TransactionDispatcher polling loop", e);
+      } finally {
+        this.pollingTimeout = setTimeout(poll, this.config.pollIntervalMs);
+        this.pollingTimeout.unref?.();
+      }
+    };
+    this.pollingTimeout = setTimeout(poll, this.config.pollIntervalMs);
```
rpanic, line 75: I didn't know about unref - that's cool!
Why is this not a interval? Because of the unref or other reasons?

packages/sequencer/src/settlement/transactions/L1TransactionRetryStrategy.ts
```diff
@@ -0,0 +1,17 @@
...
+import { PendingL1TransactionRecord } from "../../storage/repositories/PendingL1TransactionStorage";
+
+export interface L1TransactionRetryStrategy {
+  shouldRetry(record: PendingL1TransactionRecord): Promise<boolean>;
+
+  /**
+   * Optional retry delay. If provided, the dispatcher will schedule
+   * the next retry after the delay from the last sentAt time.
+   */
+  getRetryDelayMs?(record: PendingL1TransactionRecord): number;
```
rpanic, line 12: What happens if the user doesn't specify that function? When will we retry?

packages/sequencer/src/storage/repositories/PendingL1TransactionStorage.ts
```diff
@@ -0,0 +1,48 @@
...
+  | "failed";
+
+export interface PendingL1TransactionRecord {
+  id: string;
+  sender: string;
+  nonce: number;
+  attempts: number;
+  status: PendingL1TransactionStatus;
+  transaction: Transaction<any, any>;
+  hash?: string;
```
rpanic: https://github.com/o1-labs/o1js/blob/main/CHANGELOG.md#added
The PR that we requested from florian is now merged, we should use a nightly version of o1js for now until this is released but already now integrate it into this PR and use the hash everywhere

packages/sequencer/src/settlement/transactions/L1TransactionDispatcher.ts
Raunaque97: if a poll() call takes more than  pollIntervalMs time, 2 calls can overlap

packages/sequencer/src/settlement/transactions/L1TransactionRetryStrategy.ts
Raunaque97: the retry will happen after `inclusionTimeoutMs` (default: 10 mins).
this delay is after checking for status ends.

PR #370, opened 2025-12-09
Contract deployment check

### LINKED ISSUE #352: Add check that the configured settlement contract(s) are actually deployed
If the SettlementModule is started and configured with a set of contract keys - add a check that they are actually deployed on the MinaBaseLayer.
If not, we have the choice:
- Throw an error and quit the process
- Deploy the contract for them

I am leaning towards the first option

4ytuncs:
### Purpose

We needed a way to check if the keys in the settlement module configuration are indeed private keys of contracts deployed on L1. To make it, a small function is implemented, which checks the deployment via  network configuration in `MinaBaseLayer`

This PR closes #352

### Commit Descriptions

- Added token contract addresses getter: b4e2780fd5226e6be4374b4e1717a6fbf5119f62. If there are no keys, an empty array is returned.
- Deployment check function is added 0ffc741d228005ca5a1db1a30422b188817a2ca3. It is configured to check both: 
   - Checking account in the local with `hasAccount` method 
   - Checking account with `fetchAccounts` method by using network configurations given in the `Settlement.test.ts`
   
- Added test cases to check the deployment before and after deployment d9da9bd25c2c3e6ec623c251caefe59a888a885e

### DIFF
--- a/packages/sequencer/src/settlement/SettlementModule.ts
+++ b/packages/sequencer/src/settlement/SettlementModule.ts
@@ -24,6 +24,7 @@ import {
   EventEmittingComponent,
   log,
   DependencyFactory,
+  mapSequential,
 } from "@proto-kit/common";
 // eslint-disable-next-line import/no-extraneous-dependencies
 import truncate from "lodash/truncate";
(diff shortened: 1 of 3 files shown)


packages/sequencer/test/settlement/Settlement.ts
```diff
@@ -780,4 +800,18 @@ export const settlementTestFn = (
     },
     timeout
   );
+
+  it("should not throw error after settlement", async () => {
+    // If it doesn't throw anything, it indicates that deployment checs were succesful.
+    if (tokenConfig === undefined) {
+      await settlementModule.checkDeployment();
```
rpanic: Use `await expect(Promise.resolve()).resolves.toBeUndefined();`
And since this test contains a if-statement and is async, add `expect.assertions(1)` at the beginning

packages/sequencer/test/settlement/Settlement.ts
```diff
@@ -780,4 +800,18 @@ export const settlementTestFn = (
     },
     timeout
   );
+
+  it("should not throw error after settlement", async () => {
+    // If it doesn't throw anything, it indicates that deployment checs were succesful.
```
rpanic: Typo `checs`

packages/sequencer/test/settlement/Settlement.ts
```diff
@@ -578,6 +592,12 @@ export const settlementTestFn = (
           .proveAndSendTransaction(tx, "included");
 
         const actions = await Mina.fetchActions(dispatch.address);
+        if (baseLayerConfig.network.type !== "local") {
```
rpanic, line 595: We have `SettlementUtils.fetchContractAccount(accounts)` that does exactly what you did here

rpanic (changes requested):

packages/sequencer/test/settlement/Settlement.ts
4ytuncs: Done by removing it: 1707651

packages/sequencer/test/settlement/Settlement.ts
4ytuncs: My bad, done here: 1707651

packages/sequencer/test/settlement/Settlement.ts
4ytuncs: Done here: 1707651

Also, refactored the `checkDeployment` code to use less conditional checks. Passed `tokenId` regardless of its existence, since it is a conditional parameter: 4f802b5

packages/sequencer/src/settlement/SettlementModule.ts
```diff
@@ -351,4 +359,49 @@ export class SettlementModule
...
+              address: address.toBase58(),
+              error: "Not found on local chain",
+            });
+          }
+        } else {
+          const { account, error } = await fetchAccount({
+            publicKey: address,
+            tokenId,
+          });
+          if (!account) {
```
rpanic: Generally, we should not do truthy conditionals, but always make sure it ends up as a boolean.  (i thought we had a eslint rule for that)
The less javascript weirdness, the better :)

packages/sequencer/src/settlement/SettlementModule.ts
```diff
@@ -351,4 +359,49 @@ export class SettlementModule
...
+            missing.push({
+              address: address.toBase58(),
+              error: error?.statusText ?? "Not found on chain",
+            });
+          }
+        }
+      })
+    );
+
+    if (missing.length) {
```
rpanic: Same here, > 0 goes a long way here

rpanic (approved):

packages/sequencer/src/settlement/SettlementModule.ts
4ytuncs: Done: 26145ab

packages/sequencer/src/settlement/SettlementModule.ts
4ytuncs: Done: 26145ab

Also perhaps we should add a small gfi issue that detects such truthy conditionals.

PR #374, opened 2025-12-16
Refactor/make query a service

### LINKED ISSUE #349: Make Query a service
Currently, query is a manually-crafted object inside the ClientAppchain.
Lets make this a service that we resolve by class reference.
This enables us to 
1. have a cleaner separation of the implementation
2. Resolve transport module on a lazy basis, therefore not requiring the user to specify 4 different transport modules everywhere, even though they might only use one of the functionalities provided

4ytuncs:
## Description 
This PR refactors query logic in `ClientAppChain` by extracting it into a dedicated `QueryService` class with lazy initialization patterns.

## Changes 
- Added `QueryService` class with lazy-initialized getters for runtime, protocol, network, and explorer query modules f00c371
- Implemented resolver functions for transport modules `QueryTransportModule`, `NetworkStateTransportModule`, `BlockExplorerTransportModule` f00c371
- Added lazy initialization pattern to `QueryService` for query modules 23b6d4b
- Updated `ClientAppChain` to use `QueryService` 880358f

This PR closes #349

### DIFF
--- a/packages/sdk/src/client/ClientAppChain.ts
+++ b/packages/sdk/src/client/ClientAppChain.ts
@@ -19,18 +19,12 @@ import {
 } from "@proto-kit/protocol";
 import {
   DummyStateService,
-  NetworkStateQuery,
-  NetworkStateTransportModule,
   Query,
-  QueryBuilderFactory,
-  QueryTransportModule,
   Sequencer,
   UnsignedTransaction,
   AppChain,
   AppChainModule,
   MinimalAppChainDefinition,
-  BlockExplorerQuery,
-  BlockExplorerTransportModule,
 } from "@proto-kit/sequencer";
 import { container } from "tsyringe";
 import { Field, PublicKey, UInt64 } from "o1js";
@@ -42,6 +36,7 @@ import { GraphqlTransactionSender } from "../graphql/GraphqlTransactionSender";
 import { Signer } from "../transaction/InMemorySigner";
 import { AppChainTransaction } from "../transaction/AppChainTransaction";
 import { TransactionSender } from "../transaction/InMemoryTransactionSender";
+import { QueryService } from "../query/QueryService";
 
 export type InferModules<Container extends TypedClass<ModuleContainer<any>>> =
   Container extends TypedClass<infer Type>
@@ -53,6 +48,12 @@ export type InferModules<Container extends TypedClass<ModuleContainer<any>>> =
 export class ClientAppChain<
   AppChainModules extends MinimalAppChainDefinition,
 > extends AppChain<AppChainModules> {
+  // Optional for our lazy initialization purpose. 
+  private QueryService?: QueryService<
+    InferModules<AppChainModules["Runtime"]>,
+    InferModules<AppChainModules["Protocol"]>
+  >;
+
   public static from<Modules extends MinimalAppChainDefinition>(
     definition: Modules
   ) {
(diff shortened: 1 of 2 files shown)


packages/sdk/src/query/QueryService.ts
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  /**
+   * A helper function that resolves BlockExplorerTransportModule.
+   * If not resolved before, it is resolved.
+   * @returns The registered transport module as {@link BlockExplorerTransport}
+   */
+  private get blockExplorerTransport(): BlockExplorerTransportModule {
+    if (this.BlockExplorerTransport === undefined) {
+      if (!this.container.isRegistered("BlockExplorerTransportModule")) {
```
rpanic: We can save this line, tsyringe will throw a more descriptive error when it can't resolve a dependency anyways

rpanic: Same for all the others in this class

packages/sdk/src/client/ClientAppChain.ts
```diff
@@ -53,6 +48,12 @@ export type InferModules<Container extends TypedClass<ModuleContainer<any>>> =
 export class ClientAppChain<
   AppChainModules extends MinimalAppChainDefinition,
 > extends AppChain<AppChainModules> {
+  // Optional for our lazy initialization purpose.
+  private QueryService?: QueryService<
```
rpanic: lowercase

packages/sdk/src/query/QueryService.ts
```diff
@@ -0,0 +1,156 @@
...
+} from "@proto-kit/sequencer";
+
+export class QueryService<
+  RuntimeModules extends RuntimeModulesRecord,
+  ProtocolModules extends ProtocolModulesRecord &
+    MandatoryProtocolModulesRecord = ProtocolModulesRecord &
+    MandatoryProtocolModulesRecord,
+> {
+  // Here, fields are optional for lazy initialization.
+  private QueryTransport?: QueryTransportModule;
```
rpanic: So - I had a bit of a different structure in mind here - and I think this would save a lot of LOCs:
We use all the features the tsyringe gives us. 
First, we make the QueryService `@injectable()` and `@scoped(Lifecycle.ContainerScoped)`
Then, we inject al the transport module in the constructor using `@inject("Token", { isOptional: true }) private readonly module: Module | undefined`
And then, we can use this.module for all the actual queries that we have

rpanic (changes requested):

packages/sdk/src/query/QueryService.ts
rpanic: Also, in the ClientAppChain, we need to use `this.parentContainer.resolve(QueryService)`

4ytuncs: Resolved from parent container: [011abc8](https://github.com/proto-kit/framework/pull/374/commits/011abc8b478384142a7b39ef7f6ceb91699c595d)

and transport modules are now injected, so they are not directly defined in class: [3385f1d](https://github.com/proto-kit/framework/pull/374/commits/3385f1dada68c820c05be5250db2fbeabaecef61)

packages/sdk/src/query/QueryService.ts
4ytuncs: Removed since it is injected now: [3385f1d](https://github.com/proto-kit/framework/pull/374/commits/3385f1dada68c820c05be5250db2fbeabaecef61)

rpanic (approved):

PR #381, opened 2025-12-22
Feat/explorer settlement batch support

### LINKED ISSUE #147: Add support for settlement/bundles in the explorer

**Additional context**

### LINKED ISSUE #358: Revive block explorer
Currently the old block explorer has vanished from the main branch mysteriously.
In the future however, we want a modular and reusable block explorer for protokit.

We can either:
- Revive the old block explorer, update it and expand it with the new features
See https://github.com/proto-kit/framework/commit/09a32f756cf9b30a555463358da37536c9c25b0a
- Revive one of the block explorer developed during previous hackathons by community members
https://github.com/satyambnsal/ProtoScope
- Start from scratch

### LINKED ISSUE #367: Display finality status of transaction in the explorer
Currently, a transaction can have a bunch of different statuses, so we should display them accordingly in the transaction detail view.
"Pending" -> Is in the mempool
"Included" -> Is in a block
"Proven" -> Is in a batch that has been proven
"Settled" -> Settlement transaction has been sent and included
"Finalized" -> Settlement transaction has been finalized on the L1

stanlou:
Closes #147 

This PR adds  **batch and settlement support** to the explorer.

#### Indexer
- Added batch and settlement indexing and persistence in the database.
- Added GraphQL resolvers for batches and settlements.

#### Explorer UI
- Added batch list view.
- Added settlement list view.
- Refactored GraphQL queries to use variables instead of inline interpolation.
- Introduced reusable `DataTable` and shared filter components.

##### Landing Page
- Added a global search bar (blocks, transactions, settlements by hash).
- Displayed key network stats:
  - Total blocks
  - Total transactions
  - Current state root
  - Latest settlement
  - Latest block


rpanic (approved):
I have a few things that I noticed, but I think we should just add all of them as future tasks, i.e. future PRs. Just adding them here for completeness:

- Searching a block height shouldn't look for all the hashes that contain the block height, but instead I'd suggest searching for height only if the input is a number and less than the current max height (which we have access to, right?), and if the number is higher, we search for hashes

- Can we add the blocks view in the dashboard under the summaries? Kinda like https://etherscan.io/

- It would be cool if we could click on block hashes on the dashboard and get to the block's page right away

- For the transactions table inside the blocks view - can we use the same tick for the status that we use in the transaction detail page? Also, for failing transactions, lets display a x or something in red, while just adding the status message in the same column. I suspect most transactions to be successful, so the status message column would be mostly empty and it'd probably look better merged into one column.

- How hard is it to make the dashboard title and slogan configurable for users?

- Eventually, we should figure out if we want to display hashes as either hex or base64

- Also, we need to make a starter-kit PR to add the missing tasks to the module config

rpanic:
Also, lets add a changelog to the Chanelog.md

rpanic:
Also, lets add .next/ and next-env.d.ts to .gitignore as they are build artifacts

PR #388, opened 2026-01-12
Refactor/refactor provable types

### LINKED ISSUE #201: Refactor non-provable codebase to use value-types instead of provable types
In non-provable framework code (i.e. sequencer, workers, ...), we sometimes use the provable type representations (Field, PublicKey, ...) and sometimes we use the value types (bigint, string, ...)
The way o1js was refactored in the last year gives developers the right tools to only use value types and therefore achieve a better seperation between in-circuit and out-of-circuit code.

We should also use this pattern in our codebase to more clearly distinguish between the actual circuits and adjacent code. Additionally, this make our serializers leaner and saves us on conversions.

Potentially, since sometimes we use o1js-builtin functions to calculate something out of circuits (tree ops, hashing, signature verificatoin), this might also lead to some new conversions needing to happen that might offset the savings of above.
Also, we should think of how to correctly distinguish different conversations having the same js type (PublicKey and PrivateKey both encode to string).
Additionally, bigints aren't supported for json encoding yet - there are workarounds, but we should implement them framework-wide to not repeat ourselves

4ytuncs:
## Changes 

In this PR, it is aimed to convert from provable types to non-provable types in out-of-circuit code. 

Starting point changes in this refactoring was to implement or convert to JSON type of the followings: 
   - `Block`
   - `BlockWithResult`

And since they are 
   - `TransactionExecutionResult`
   - `PendingTransaction`
   - `StateTransitionBatch`
   - `UntypedStateTransition`
   - `NetworkState`

Especially last 3 is used all around the Sequencing and Tracing modules. 

#### UntypedStateTransition & StateTransitionBatch
- Implement `UntypedStateTransitionJson` 0199877 ( rename:  5934a13 )
- Implement in block and tracing services 0199877
- Change in tracing, block result services 5934a13

#### TransactionExecutionResult
- Implemented Json interface, `TransactionExecutionResultJson`  b017bb5 
- Implement conversion function between json type and provable type to use when needed 08f95bb
- Changed in execution and tracing 564ba7a
- Changed in Prisma storage 6905b36
- Changed in various modules fb4c854

#### PendingTransaction
- Changed in `PrivateMempool` d3c9336
- Changed in various modules  d5b6d7b 
- Changed in storage and mempool d8558f9 
- Changed in execution and tracing 564ba7a

#### NetworkState
- Define `NetworkStateJson` by inferring it from Struct c29a51d 
- add `RuntimeProofParametersJson` type with `NetworkStateJson` and `PendingTransactionJsonType` a4ac392
- 

#### Miscellanous
- refactor CllientBlock with value types 1093afe

#### Test
- Updated changes in test files 57ad69d

This PR closes #201


4ytuncs:
I've updated the commits - there might be missing parts in codebase, but most of the changes will be done in this manner. 

I'd like to have your opinion on a decision: 

The fields of `Block` and `BlockWithResult` are modified directly, whereas interfaces such as `TransactionWithResult` and `StateTransitionBatch` are neither removed nor changed. This can be discussed, but i am for keeping both provable and non provable types, since having them enables us the access provable types in an easier way at the boundaries of the provable code.

packages/protocol/src/utils/utils.ts
```diff
@@ -33,6 +35,14 @@ export function notInCircuit(): MethodDecorator {
   };
 }
 
+export function networkStateToFields(json: NetworkStateJson): Field[] {
```
rpanic: This function shouldn't exist - or at least the implementation should be NetworkState.toFields(NetworkState.fromJSON())

packages/protocol/src/utils/utils.ts
```diff
@@ -33,6 +35,14 @@ export function notInCircuit(): MethodDecorator {
...
+  return [Field(json.block.height), Field(json.previous.rootHash)];
+}
+
+export function hashNetworkState(json: NetworkStateJson): string {
```
rpanic: Same here, we already implement hash() in the provable version, so we should use this to not run into inconsistencies

packages/sdk/src/graphql/GraphqlBlockExplorerTransportModule.ts
```diff
@@ -106,14 +105,14 @@ export class GraphqlBlockExplorerTransportModule
...
       blockData.previousBlockHash != null && blockData.previousBlockHash !== ""
-        ? Field(blockData.previousBlockHash)
+        ? blockData.previousBlockHash
         : undefined;
 
     return {
-      hash: Field(blockData.hash),
-      height: Field(blockData.height),
+      hash: blockData.hash,
+      height: blockData.height.toString(),
```
rpanic: height is number and can stay number imo - no need to convert into strings here

packages/sequencer/src/mempool/PendingTransaction.ts
```diff
@@ -118,7 +118,7 @@ export class UnsignedTransaction implements UnsignedTransactionBody {
   }
 }
 
-interface PendingTransactionJSONType {
+export interface PendingTransactionJSONType {
```
rpanic: So here, since PendingTransaction is never used inside circuits, we can just get rid of this type, and instead make PendingTransaction non-provable. That can get tricky with the signatures - we can keep UnsignedTransaction provable for now? Or create a new transaction creation API altogether

packages/sdk/src/graphql/GraphqlNetworkStateTransportModule.ts
```diff
@@ -3,7 +3,7 @@ import {
   NetworkStateTransportModule,
   AppChainModule,
 } from "@proto-kit/sequencer";
-import { NetworkState } from "@proto-kit/protocol";
+import { NetworkStateJson } from "@proto-kit/protocol";
```
rpanic: I'm a bit unsure, is the naming (NetworkState, NetworkStateJson) or (ProvableNetworkState, NetworkState) better?

packages/sdk/src/query/InMemoryBlockExplorer.ts
```diff
@@ -64,27 +64,26 @@ export class InMemoryBlockExplorer
     const clientTransactions: ClientTransaction[] = block.transactions.map(
       (txResult) => ({
         tx: {
-          hash: txResult.tx.hash().toString(),
-          methodId: txResult.tx.methodId.toString(),
-          nonce: txResult.tx.nonce.toString(),
-          sender: txResult.tx.sender.toBase58(),
-          argsFields: txResult.tx.argsFields.map((f) => f.toString()),
+          hash: txResult.tx.hash,
```
rpanic: We should just return txResult here :))

packages/sdk/src/transaction/AppChainTransaction.ts
```diff
@@ -47,6 +47,7 @@ export class AppChainTransaction {
 
   public async send() {
     this.hasPendingTransaction(this.transaction);
-    await this.transactionSender.send(this.transaction);
+    // Wait for next commit
```
rpanic: marker to resolve this

packages/sequencer/src/helpers/utils.ts
```diff
@@ -194,3 +194,17 @@ export class PairProofTaskSerializer<PublicInputType, PublicOutputType>
     );
   }
 }
+
+/**
+ * Type for serialized {@link Field}.
+ */
+export type FieldString = string;
```
rpanic: Let's move this to common/utils

packages/sequencer/src/storage/model/Block.ts
```diff
@@ -28,28 +37,46 @@ export interface TransactionExecutionResult {
...
+  hooksStatus: boolean;
+  statusMessage?: string;
+  events: {
+    eventName: string;
+    data: FieldString[];
+    source: "afterTxHook" | "beforeTxHook" | "runtime";
+  }[];
+}
+
 // TODO Why is Block using Fields, but BlockResult bigints? Align that towards the best option
```
rpanic: Remove this comment since it's solved with this PR

packages/sequencer/src/storage/model/Block.ts
```diff
@@ -28,28 +37,46 @@ export interface TransactionExecutionResult {
   }[];
 }
 
+export interface StateTransitionBatchJson {
```
rpanic: The the StateTransitionBatch and TxExecutionResult are only used inside the sequencer - we shouldn't make two slighty different copies, but instead change the existing types and adapt everything from there

packages/sequencer/src/storage/model/Block.ts
```diff
@@ -59,18 +86,21 @@ export const Block = {
...
 
 export interface BlockResult {
-  blockHash: bigint;
-  witnessedRoots: [bigint];
-  stateRoot: bigint;
-  blockHashRoot: bigint;
-  afterNetworkState: NetworkState;
-  afterBlockStateTransitions: UntypedStateTransition[];
-  blockHashWitness: BlockHashMerkleTreeWitness;
+  blockHash: string;
```
rpanic: Why string here and not FieldString??

packages/sequencer/src/protocol/production/helpers/UntypedStateTransition.ts
```diff
@@ -3,6 +3,20 @@ import { ProvableStateTransition, StateTransition } from "@proto-kit/protocol";
 
 import { UntypedOption } from "./UntypedOption";
 
+export interface UntypedStateTransitionJson {
```
rpanic: Here as well - we exclusively use the original UntypedStateTransition inside the sequencer, therefore there's no reason to have two variants of the same thing, lets just change the old one

packages/sequencer/src/protocol/production/sequencing/BlockProductionService.ts
```diff
@@ -164,33 +169,37 @@ export class BlockProductionService {
...
       fromEternalTransactionsHash: lastBlock.toEternalTransactionsHash,
-      toEternalTransactionsHash:
-        newBlockState.eternalTransactionsList.commitment,
+      toEternalTransactionsHash: FieldString(
+        newBlockState.eternalTransactionsList.commitment
+      ),
       height:
-        lastBlock.hash.toBigInt() !== 0n ? lastBlock.height.add(1) : Field(0),
-      fromBlockHashRoot: Field(lastResult.blockHashRoot),
+        FieldString(lastBlock.hash) !== "0"
```
rpanic: First of all, lastBlock is already value-typed, so we don't have to wrap everything. Then, why cast the height to a bigint when its a number in the first place? height is number everywhere, so no use in making it a bigint

packages/sequencer/src/protocol/production/sequencing/BlockResultService.ts
```diff
@@ -218,11 +224,11 @@ export class BlockResultService {
       {
         blockHashRoot,
         stateRoot: witnessedStateRoot,
-        incomingMessagesHash: block.toMessagesHash,
-        transactionsHash: block.transactionsHash,
-        eternalTransactionsHash: block.toEternalTransactionsHash,
+        incomingMessagesHash: Field(block.toMessagesHash),
```
rpanic, line 220: Note to myself: We should check if we can pass in Value objects into circuits directly (I know this works with witnesses, but not sure about zkprogram args)

packages/sequencer/src/protocol/production/sequencing/TransactionExecutionService.ts
```diff
@@ -103,7 +107,10 @@ async function decodeTransaction(
     module,
     methodName
   );
-  const args = await parameterDecoder.decode(tx.argsFields, tx.auxiliaryData);
+  const args = await parameterDecoder.decode(
```
rpanic: Lets adjust the decoder to take strings instead of mapping here

packages/sequencer/src/protocol/production/sequencing/TransactionExecutionService.ts
```diff
@@ -539,15 +548,21 @@ export class TransactionExecutionService {
...
           beforeTxHookResult.statusMessage ??
           afterTxHookResult.statusMessage ??
           runtimeResult.statusMessage,
 
-        stateTransitions,
-        events: beforeHookEvents.concat(runtimeResultEvents, afterHookEvents),
+        stateTransitions: stateTransitions.map(STBatchToJson),
+        events: beforeHookEvents
+          .concat(runtimeResultEvents, afterHookEvents)
+          .map((e) => ({
```
rpanic, line 551: No inline mappers! Let's chat async what is best to do here

packages/sequencer/src/protocol/production/tasks/RuntimeProvingTask.ts
```diff
@@ -29,11 +37,17 @@ export interface RuntimeProofParameters {
   state: TaskStateRecord;
 }
 
+export interface RuntimeProofParametersJson {
```
rpanic: yuck - change the existing type please :)

packages/sequencer/src/protocol/production/tasks/serializers/NewBlockProvingParametersSerializer.ts
```diff
@@ -71,14 +67,9 @@ export class NewBlockProvingParametersSerializer
...
-        startingStateBeforeHook: DecodedStateSerializer.toJSON(
-          input.params.startingStateBeforeHook
-        ),
-
-        startingStateAfterHook: DecodedStateSerializer.toJSON(
-          input.params.startingStateAfterHook
-        ),
+        startingStateBeforeHook: input.params.startingStateBeforeHook,
 
+        startingStateAfterHook: input.params.startingStateAfterHook,
```
rpanic: Insert empty line

packages/sequencer/src/protocol/production/tasks/serializers/RuntimeProofParametersSerializer.ts
```diff
@@ -1,41 +1,19 @@
...
-
-import {
-  DecodedStateSerializer,
-  JSONEncodableState,
-} from "./DecodedStateSerializer";
+import type { RuntimeProofParametersJson } from "../RuntimeProvingTask";
 
+/**
+ * Serializer for RuntimeProofParametersJson.
+ * Since RuntimeProofParametersJson is already JSON-compatible, this is trivial.
```
rpanic: Indeed this is trivial, hence why we have a JsonTaskSerializer that does this already - let's remove this one and replace usage with the JsonTaskSerializer

packages/sequencer/src/protocol/production/tracing/BlockTracingService.ts
```diff
@@ -23,6 +29,30 @@ import {
 
 export type TaskStateRecord = Record<string, Field[]>;
 
+export type TaskStateRecordJson = Record<string, string[]>;
```
rpanic: again - replace the existing struct

packages/sequencer/src/protocol/production/tracing/BlockTracingService.ts
```diff
@@ -23,6 +29,30 @@ import {
...
+
+export function taskStateRecordToJson(
```
rpanic: This should never live in a component like this - either implement a TaskSerializer or change the methods that create TaskStateRecords to do the serialization

packages/sequencer/src/protocol/production/tracing/TransactionTracingService.ts
```diff
@@ -113,26 +124,27 @@ export class TransactionTracingService {
...
   private createRuntimeProofParams(
-    tx: TransactionExecutionResult,
+    tx: TransactionExecutionResultJson,
     networkState: NetworkState
-  ): RuntimeProofParameters {
-    const startingState = collectStartingState(
-      tx.stateTransitions[1].stateTransitions
-    );
+  ): RuntimeProofParametersJson {
+    const stBatch = STBatchFromJson(tx.stateTransitions[1]);
```
rpanic: Just seeing this - again, no standalone mapping functions, and then the naming case...

rpanic (changes requested):

packages/sequencer/src/mempool/PendingTransaction.ts
```diff
@@ -27,6 +27,106 @@ export type UnsignedTransactionBody = {
...
+  isMessage: boolean;
+}
+
+export class PendingTransaction {
+  public readonly hash: string;
+
+  public readonly methodId: string;
+
+  public readonly nonce: string;
+
```
4ytuncs, line 50: Moved it up to fix linting issues.

packages/sequencer/src/protocol/production/tasks/serializers/RuntimeProofParametersSerializer.ts
4ytuncs: It is removed: [81c0df7](https://github.com/proto-kit/framework/pull/388/commits/81c0df7646c2d51d00d4fb53568a45c46065bbd6)

packages/sequencer/src/protocol/production/sequencing/TransactionExecutionService.ts
4ytuncs: Event type, in RuntimeMethodExecutionContext, is declared as `any` https://github.com/proto-kit/framework/blob/e4b9e3c601583d5ef63f90dee832ba8a5c8085f8/packages/protocol/src/state/context/RuntimeMethodExecutionContext.ts#L30

I assume it is coming in a type that can be converted to the `String` directly, since it can be converted to `Field` with `toFields`

packages/sequencer/src/protocol/production/sequencing/BlockProducerModule.ts
```diff
@@ -101,7 +96,7 @@ export class BlockProducerModule extends SequencerModule<BlockConfig> {
         );
       }
       log.info(
-        `Status: ${tx.status.toBoolean()}`,
+        `Status: ${tx.status ? "true" : "false"}`,
```
rpanic: Remove ternary here and use boolean directly in template string

packages/sequencer/src/protocol/production/tracing/BlockTracingService.ts
```diff
@@ -50,12 +50,14 @@ export class BlockTracingService {
     "publicInput" | "networkState" | "blockWitness"
   > {
     const publicInput: BlockProverPublicInput = new BlockProverPublicInput({
-      stateRoot: state.stateRoot,
-      blockNumber: firstBlock.height,
-      blockHashRoot: firstBlock.fromBlockHashRoot,
-      eternalTransactionsHash: firstBlock.fromEternalTransactionsHash,
-      incomingMessagesHash: firstBlock.fromMessagesHash,
-      networkStateHash: firstBlock.networkState.before.hash(),
+      stateRoot: Field(state.stateRoot),
```
rpanic: Remove unnecessary serializations here

packages/sequencer/src/protocol/production/helpers/UntypedOption.ts
```diff
@@ -22,26 +32,34 @@ export class UntypedOption extends OptionBase {
...
+    return new UntypedOption(this.isSome, [...this.value], this.isForcedSome);
+  }
+
+  public forceSome() {
+    this.isForcedSome = !this.isSome;
+    this.isSome = true;
+  }
+
+  public encodeValueToFields(): Field[] {
+    return this.value.map((fieldString) => Field(fieldString));
```
rpanic: Can we make this return string[] and then in line 13 we convert to Fields explicity (versus the other way around how you do it currently)

packages/sequencer/src/protocol/production/helpers/UntypedOption.ts
```diff
@@ -22,26 +32,34 @@ export class UntypedOption extends OptionBase {
...
   }
 
-  protected encodeValueToFields(): Field[] {
-    return this.value;
+  public toProvable(): ProvableOption {
+    return new ProvableOption({
```
rpanic: We should be able to use ProvableOption.fromJSON() here

packages/sequencer/src/protocol/production/helpers/UntypedStateTransition.ts
```diff
@@ -26,45 +22,34 @@ export class UntypedStateTransition {
...
-   * while enforcing the 'from' property to be 'Some' in all cases.
-   */
   public toProvable(): ProvableStateTransition {
+    const from = this.from.clone();
+    from.forceSome();
+
     return new ProvableStateTransition({
-      path: this.path,
-      from: this.from.toProvable(),
+      path: Field(this.path),
```
rpanic, line 57: Let's try using ProvableStateTransition.fromJSON here as well - not sure if it works though (because of the nested struct)

packages/sequencer/src/protocol/production/helpers/UntypedStateTransition.ts
```diff
@@ -26,45 +22,34 @@ export class UntypedStateTransition {
...
+      to: this.to.toJSON(),
+    };
   }
 
-  /**
-   * Converts a StateTransition to a ProvableStateTransition,
-   * while enforcing the 'from' property to be 'Some' in all cases.
-   */
   public toProvable(): ProvableStateTransition {
+    const from = this.from.clone();
```
rpanic: Those two lines should just be 
this.from
The getter already takes care of everything you do here

packages/sequencer/src/protocol/production/helpers/UntypedStateTransition.ts
```diff
@@ -26,45 +22,34 @@ export class UntypedStateTransition {
...
-    from.forceSome();
-    return from;
-  }
-
-  public get to() {
-    return this.toValue.clone();
+  public toJSON() {
+    return {
+      path: this.path,
+      from: this.from.toJSON(),
```
rpanic: We have to use .fromValue here, otherwise this function will mutate the from.isSome, which for serialization alone, we don't want

packages/sequencer/src/protocol/production/helpers/UntypedStateTransition.ts
```diff
@@ -26,45 +22,34 @@ export class UntypedStateTransition {
...
   public constructor(
-    public path: Field,
-    public fromValue: UntypedOption,
-    public toValue: UntypedOption
+    public path: string,
+    public from: UntypedOption,
+    public to: UntypedOption
   ) {}
 
-  public get from() {
```
rpanic, line 41: Don't remove the getters here, they have a purpose :))

packages/sequencer/src/protocol/production/sequencing/BlockResultService.ts
```diff
@@ -36,15 +37,12 @@ import { executeWithExecutionContext } from "./TransactionExecutionService";
...
-    (state, st) => {
-      if (st.toValue.isSome.toBoolean()) {
-        state[st.path.toString()] = st.toValue.value;
-      }
-      return state;
-    },
-    {}
-  );
+  return stateTransitions.reduce<StateRecord>((state, st) => {
+    if (st.to.isSome) {
```
rpanic: Change back to toValue

packages/sequencer/src/protocol/production/sequencing/BlockResultService.ts
```diff
@@ -56,7 +54,9 @@ function createCombinedOrderedStateDiff(
     .map((tx) => {
       const transitions = tx.stateTransitions
         .filter(({ applied }) => applied)
-        .flatMap(({ stateTransitions }) => stateTransitions);
+        .flatMap(({ stateTransitions }) =>
+          stateTransitions.map((st) => UntypedStateTransition.fromJSON(st))
```
rpanic: This shouldn't be here, you can just use the stateTransitions directly

packages/indexer/src/IndexerNotifier.ts
```diff
@@ -69,6 +66,7 @@ export class IndexerNotifier extends SequencerModule<Record<never, never>> {
         const txQueue = await this.taskQueue.getQueue(
           this.indexPendingTxTask.name
         );
+        // This part seems weird
```
rpanic, line 69: Remove this comment

packages/indexer/test/IndexerNotifier.test.ts
```diff
@@ -130,7 +132,7 @@ async function sendTransactions(
       { nonce: i }
     );
 
-    console.log("tx nonce", tx.transaction?.nonce.toBigInt());
+    console.log("tx nonce", (tx.transaction as PendingTransaction).nonce);
```
rpanic, line 135: Why do you do this? tx should already be of type PendingTransaction, no?

packages/persistance/src/services/prisma/mappers/EventMapper.ts
```diff
@@ -1,31 +1,30 @@
 import { singleton } from "tsyringe";
 import { Prisma } from "@prisma/client";
-import { Field } from "o1js";
 
 import { ObjectMapper } from "../../../ObjectMapper";
 
-type EventData = {
+type EventDataJson = {
```
rpanic, line 6: I think we should leave this as EventData - there's not provable version of it, so that name is fine imo

packages/persistance/test-integration/PrismaBlockProduction.test.ts
```diff
@@ -239,10 +224,16 @@ describe("prisma integration", () => {
 
       expectDefined(transaction.transaction);
 
+      let txHash: string;
+
+      if (transaction.transaction instanceof PendingTransaction) {
```
rpanic, line 229: Whaat - why do we have this conditional in a test?

packages/processor/src/triggers/TimedProcessorTrigger.ts
```diff
@@ -40,19 +40,19 @@ export class TimedProcessorTrigger extends ProcessorModule<TimedProcessorTrigger
       return false;
     }
 
-    if (lastProcessedBlockHeight === Number(block?.block.height.toBigInt())) {
+    if (lastProcessedBlockHeight === Number(block?.block.height)) {
```
rpanic, line 43: Can't we skip this cast to number since it's already a number?

packages/processor/test/HandlersExecutor.test.ts
```diff
@@ -87,7 +87,7 @@ describe("HandlersModule", () => {
           await client.balance.create({
             data: {
               address: from.toBase58(),
-              height: Number(block.height.toString()),
+              height: Number(block.height),
```
rpanic, line 90: Same here, number already

packages/sdk/src/graphql/GraphqlNetworkStateTransportModule.ts
```diff
@@ -30,7 +30,7 @@ export class GraphqlNetworkStateTransportModule
 
   private async retrieveNetworkState(path: string): Promise<NetworkState> {
     const query = gql`
-      query NetworkState {
+      query ProvableNetworkState {
```
rpanic, line 33: Yeah this is probably a wrong renaming that your IDE made

packages/sdk/src/graphql/GraphqlTransactionSender.ts
```diff
@@ -23,7 +23,7 @@ export class GraphqlTransactionSender
         submitTx(tx: $tx)
       }
     `;
-    const tx = transaction.toJSON();
+    const tx = transaction;
```
rpanic, line 26: yeah that line became a bit nonsensical

packages/sequencer/src/mempool/PendingTransaction.ts
```diff
@@ -27,6 +27,106 @@ export type UnsignedTransactionBody = {
   isMessage: boolean;
 };
 
+interface PendingTransactionJSONType {
```
rpanic, line 30: That type became unnecessary (since its the same as PendingTransaction)

packages/sequencer/src/mempool/PendingTransaction.ts
```diff
@@ -27,6 +27,106 @@ export type UnsignedTransactionBody = {
...
+
+  public readonly nonce: string;
+
+  public readonly sender: string;
+
+  public readonly argsFields: string[];
+
+  public readonly auxiliaryData: string[];
+
+  public readonly signature: { r: string; s: string };
```
rpanic, line 57: Maybe there's some signature json type that we can use here?

packages/sequencer/src/mempool/PendingTransaction.ts
```diff
@@ -27,6 +27,106 @@ export type UnsignedTransactionBody = {
...
+  argsFields: string[];
+  auxiliaryData: string[];
+  signature: {
+    r: string;
+    s: string;
+  };
+  isMessage: boolean;
+}
+
+export class PendingTransaction {
```
rpanic, line 44: Why did you remove the inheritance of UnsignedTransaction here? UnsignedTransaction is only used for user-facing APIs and there was a reason I built it this way. That's so we can sign nicely, verify signatures nicely, etc.
No reason to strip that away from the API

packages/sequencer/src/mempool/verification/TransactionValidator.ts
```diff
@@ -45,10 +46,14 @@ export class TransactionValidator {
       return [false, methodError];
     }
 
-    const validSignature = tx.signature.verify(
-      tx.sender,
-      tx.getSignatureData()
-    );
+    const signature = Signature.fromJSON(tx.signature);
```
rpanic, line 49: This relates to the comment in PendingTransaction - no reason why we would embedd this here, this makes the codebase very fragile if we don't take those pure functions and make them reusable

packages/sequencer/src/protocol/production/helpers/UntypedOption.ts
```diff
@@ -1,15 +1,27 @@
...
+export class UntypedOption {
+  public constructor(
+    public isSome: boolean,
+    public value: string[],
+    public isForcedSome: boolean
+  ) {}
+
+  public get treeValue(): string {
+    if (this.isSome && !this.isForcedSome) {
+      return Poseidon.hash(this.encodeValueToFields()
```
rpanic: Formatting can't be right here

packages/sequencer/src/protocol/production/helpers/UntypedOption.ts
```diff
@@ -1,15 +1,27 @@
...
+import { Option, ProvableOption } from "@proto-kit/protocol";
+
+export class UntypedOption {
+  public constructor(
+    public isSome: boolean,
+    public value: string[],
+    public isForcedSome: boolean
+  ) {}
+
+  public get treeValue(): string {
```
rpanic, line 11: Also, if you have time, we should add a equivalence test here checking that UntypedOption.treeValue returns the same as Option.treeValue()

packages/sequencer/src/protocol/production/helpers/UntypedOption.ts
```diff
@@ -1,15 +1,27 @@
...
-/**
- * Option facilitating in-circuit values that may or may not exist.
- */
-export class UntypedOption extends OptionBase {
   public static fromOption<Value>(option: Option<Value> | Option<Field>) {
     return new UntypedOption(
-      option.isSome,
-      option.encodeValueToFields(),
-      option.isForcedSome
+      option.isSome.toBoolean(),
```
rpanic, line 22: Wouldn't we be able to use Option.toJSON here?

rpanic: Oh okay, It'd return a object, but we should still use it and then deconstruct the returned object

packages/sequencer/src/protocol/production/helpers/UntypedOption.ts
```diff
@@ -22,26 +34,36 @@ export class UntypedOption extends OptionBase {
...
+    return {
+      isSome: this.isSome,
+      value: this.value,
+      isForcedSome: this.isForcedSome,
+    };
   }
 
   public clone() {
-    return new UntypedOption(this.isSome, this.value, this.isForcedSome);
+    return new UntypedOption(this.isSome, [...this.value], this.isForcedSome);
```
rpanic, line 49: use this.value.slice() here - this will also make a shallow copy and is more explicit

packages/sequencer/src/protocol/production/helpers/UntypedStateTransition.ts
```diff
@@ -26,14 +22,14 @@ export class UntypedStateTransition {
...
-      Field(path),
+      path,
       UntypedOption.fromJSON(from),
       UntypedOption.fromJSON(to)
     );
   }
 
   public constructor(
-    public path: Field,
+    public path: string,
```
rpanic, line 32: FieldString

packages/sequencer/src/protocol/production/tasks/RuntimeProvingTask.ts
```diff
@@ -49,22 +53,22 @@ export class RuntimeProvingTask
   }
 
   public inputSerializer(): TaskSerializer<RuntimeProofParameters> {
-    return new RuntimeProofParametersSerializer();
+    return JSONTaskSerializer.fromType<RuntimeProofParameters>();
```
rpanic, line 56: Okay, so I get why you did this here, but you ended up having to reimplement all the logic of serialization that that serializer encapsulated, so this change is very bad for the code quality imo. 
The reason why the TaskSerializers exist is exactly to move the parsing logic away from the compute() function to a separate, isolated component

packages/sequencer/src/protocol/production/tasks/TransactionProvingTask.ts
```diff
@@ -34,7 +37,7 @@ export async function executeWithPrefilledStateService<Return>(
     .forEach((startingState) => {
       stateServiceProvider.setCurrentStateService(
         new PreFilledStateService({
-          ...startingState,
+          ...DecodedStateSerializer.fromJSON(startingState),
```
rpanic, line 40: Imo this goes into the same category of "not serializing where logic happens" - so while the total LOC goes down with this change, it increases mental complexity and decreases separation of concerns

packages/sequencer/src/protocol/production/tasks/serializers/types/TransactionProvingTypes.ts
```diff
@@ -5,14 +5,14 @@ import {
...
-import type { TaskStateRecord } from "../../../tracing/BlockTracingService";
+import { JSONEncodableState } from "../DecodedStateSerializer";
 
 export type RuntimeProof = Proof<void, MethodPublicOutput>;
 
 export interface TransactionProverTaskParameters {
   publicInput: TransactionProverPublicInput;
   executionData: TransactionProverExecutionData;
-  startingState: TaskStateRecord[];
+  startingState: JSONEncodableState[];
```
rpanic, line 15: Same thing here :)

packages/sequencer/src/protocol/production/tracing/StateTransitionTracingService.ts
```diff
@@ -50,7 +50,12 @@ export class StateTransitionTracingService {
           stateTransitions: block.block.beforeBlockStateTransitions,
           applied: true,
         },
-        ...block.block.transactions.flatMap((tx) => tx.stateTransitions),
+        ...block.block.transactions.flatMap((tx) =>
```
rpanic, line 53: Whaaat - this is semantically the same, why did you change this?

packages/sequencer/src/mempool/PendingTransaction.ts
```diff
@@ -27,6 +27,106 @@ export type UnsignedTransactionBody = {
...
+  public readonly hash: string;
```
rpanic, line 45: Any particular reason why you made the hash a field here? We memoized it previously, but it was always a computation based on all other fields of the transactions - having to compute the hash manually before creating this object is a non-starter imo. See UnsignedTransaction on how it was implemented previously (we should reuse that implementation like i mentioned above)

packages/sequencer/src/settlement/messages/MinaIncomingMessageAdapter.ts
```diff
@@ -87,13 +87,22 @@ export class MinaIncomingMessageAdapter implements IncomingMessageAdapter {
     const args = await methodEncoder.decode(fieldArgs, []);
 
     const { fields, auxiliary } = methodEncoder.encode(args);
+    const hash = Poseidon.hash([
```
rpanic, line 90: Like I mentioned in PendingTransaction.ts, we shouldn't do this!

packages/sequencer/test/integration/BlockProduction-test.ts
```diff
@@ -231,8 +232,20 @@ export function testBlockProduction<
     expectDefined(latestBlockWithResult);
     expectDefined(latestBlockWithResult.result);
     expect(
-      latestBlockWithResult.result.afterNetworkState.hash().toString()
-    ).toStrictEqual(batch!.toNetworkState.hash().toString());
+      new ProvableNetworkState(
+        ProvableNetworkState.fromJSON(
```
rpanic, line 236: We have these 5 lines at least 20 times in the codebase - should we make NetworkState a class and implement toProvable and fromProvable on it?

packages/sequencer/src/mempool/PendingTransaction.ts
```diff
@@ -27,6 +27,106 @@ export type UnsignedTransactionBody = {
   isMessage: boolean;
 };
 
+interface PendingTransactionJSONType {
+  hash: string;
+  methodId: string;
+  nonce: string;
```
rpanic, line 33: Why is nonce a string? It should be a number

packages/sequencer/test/settlement/Settlement-only.ts
```diff
@@ -155,6 +156,7 @@ export const settlementOnlyTestFn = (
   async function createBatch(
     withTransactions: boolean,
     customNonce: number = 0,
+    // Why is it like this?
```
rpanic, line 159: What are you referring to?

packages/stack/test/graphql/graphql.test.ts
```diff
@@ -184,7 +185,10 @@ describe("graphql client test", () => {
     await tx.sign();
     await tx.send();
 
-    const txHash = tx.transaction?.hash().toString()!;
+    const txHash =
+      tx.transaction instanceof UnsignedTransaction
```
rpanic, line 189: Yeah no, we need to fix the PendingTransaction implementation :)

packages/indexer/src/IndexerNotifier.ts
4ytuncs: Done in the previous commits, so resolving from here

packages/sequencer/test/integration/BlockProduction-test.ts
4ytuncs: Not sure - even though toProvable and fromProvable will be good to remove these parts, we could need fromJson and toJson in that class too. 

Maybe not though, have to check where `NetworkState` is used.

packages/sequencer/test/settlement/Settlement-only.ts
4ytuncs: A note for myself that where i thought of using PendingTransaction as json, this line will be removed.

packages/sequencer/src/protocol/production/tracing/StateTransitionTracingService.ts
4ytuncs: Must have left from the PendingTransaction's JSON era.

rpanic:
Closed as not planned anytime soon

PR #392, opened 2026-01-14
S3 prover artifact caching

### LINKED ISSUE #391: Implement a remote caching solution for circuit compilation
Since in a production setup, we will have a large set of workers, we want them to be able to access a cache such that they don't have to each compile all the circuits every time they get spun up.
Currently, o1js only allows for synchronous caching (with a implementation for storage in the file system). However, we can build a mechanism around it to on-demand fetch artifacts from a S3 bucket to the store on the file system, so that o1js can then read it from the file system.

rpanic:
Closes #391 

Based on #380


packages/stack/test-integration/caching/caching.test.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    Protocol: Protocol.from(VanillaProtocolModules.with({})),
+
+    Sequencer: Sequencer.from({
+      Database: InMemoryDatabase,
+
+      Mempool: PrivateMempool,
+      LocalTaskWorkerModule: LocalTaskWorkerModule.from(
+        VanillaTaskWorkerModules.withoutSettlement()
+      ),
```
4ytuncs, line 54: Can't we use `DefaultModules` & `DefaultConfigs` here?

packages/sequencer/test-integration/workers/workers-proven.test.ts
```diff
@@ -20,13 +20,15 @@ import { BlockTestService } from "../../test/integration/services/BlockTestServi
...
   protocolClass,
+  RemoteCacheConfig,
   runtimeClass,
   runtimeProtocolConfig,
 } from "./modules";
 import { ChildProcessWorker } from "./ChildProcessWorker";
 
 const timeout = 300000;
 
+// true
```
4ytuncs, line 31: Why is this here

4ytuncs (approved):
Left two comments, probably would not require changes.

4ytuncs (dismissed):
Left comments about some very minor things.

4ytuncs (approved):

PR #398, opened 2026-01-20
feat: add default modules and configurations for sequencer presets

stanlou:
## Summary

This PR introduces new classes in the  **`stack`** package that provide ready-to-use defaults for modules and configuration.

### Default Modules & Configs
- **`DefaultModules`** and **`DefaultConfigs`**
  - Centralized helpers that return the default module set and configuration for each supported option.
  - Designed to be composable and easy to override.

## Goal

The goal of this change is to:
- Reduce boilerplate for common setups
- Provide clear defaults
- Make it easy to bootstrap a Sequencer or AppChain while still allowing customization and overrides


packages/stack/src/presets/config.ts
```diff
@@ -0,0 +1,184 @@
+export const inmemoryConfig = {
+  blockInterval: 5000,
+  graphqlHost: "localhost",
+  graphqlPort: 8080,
+  graphiqlEnabled: true,
+};
+export const developmentConfig = {
+  proofsEnabled: false,
```
rpanic, line 8: I really don't like that we are redefining all these configuration options here. Because we already have them defined on a module level - ConfigurableModule<Config> allows us to extract config typing.
I would queue a followup PR where we try to remove this middle layer and reuse the config types directly. Obviously we need some translation layer from ENV variables -> configs - we can think about how to best do this.
We should also separate clearly between mandatory configs, which we *have* to set in the starter kit (things like ports, connection strings, etc. - these will mostly be env variables as well) and then non-mandatory configuration, of which we have a quite a few as well, but which aren't included in here at all yet - that's where reusing the module's config types comes in handy

packages/stack/src/presets/modules/types.ts
```diff
@@ -0,0 +1,63 @@
...
+  graphqlHost: string;
+  graphiqlEnabled: boolean;
+};
+export type CoreEnv = {
+  blockInterval: number;
+  settlementInterval?: number;
+  minaBridgeContractPrivateKey?: string;
+  customTokenPrivateKey?: string;
+  customTokenBridgePrivateKey?: string;
+};
```
rpanic, line 14: There should be newlines in here - eslint doesn't complain about that?

rpanic: It doesn't - creating a PR enforcing that, sorry :)))

packages/stack/src/presets/modules/index.ts
```diff
@@ -0,0 +1,518 @@
...
+    } satisfies SequencerModulesRecord;
+  }
+
+  static inMemoryDatabase() {
+    return {
+      Database: InMemoryDatabase,
+    } satisfies SequencerModulesRecord;
+  }
+
+  static PrismaRedisDatabase() {
```
rpanic, line 147: Why is this upper case and most other things are lower case? Same for RedisTaskQueue

rpanic (approved):

PR #409, opened 2026-01-28
Feat/cli extended functionality

### LINKED ISSUE #353: New starter kit structure (v3)
The current starter kit is very bulky and unflexible to use, customize and configure. So we should revamp the structure a bit to make it easier to understand and use.
A few things that I've noticed:

1. We offer a typesafe way to configure appchains, however the starter kit uses environment variables to set those values, leading to a ton of variables and a lot of overhead since those need to be propagated to the actual configuration API is a cumbersome way.
Some solutions:
Allow the Appchain configure API to consume env variables directly
Remove env variables entirely (with the exception of docker-related stuff) and use configurations directly

2. We toggle settlement functionality using a conditional modules record creation - we should make this cleaner (and introduce typesafety back in)

The general direction I want to move in:
Separate the modules definition and the configuration values in a way that can be composed flexibly and directly with each other.
Use the CLI for this composition and in addition to that, use it for script execution and environment selection

Additional ideas:
- [ ] Separate runtime + protokit from sequencer and environment (mainly in terms of folder composition
- [ ] Ship most module configurations from the framework -> reduces the starter-kit size and additionally improved upgradability
- [ ] Add "environment wizard" for users to easily scaffold their own environments, go through all the configuration options and let users choose - then generate a module config based on their choices

stanlou:
## Description
This PR extends the Proto-Kit CLI with new commands for environment setup, bridge , settlement deployment, and explorer management. 

### Environment Management

**`wizard`**
- Creates a new environment configuration with an interactive guided wizard
- Example: `protokit wizard`

### Bridge Operations

**`bridge deposit <tokenId> <fromKey> <toKey> <amount>`**
- Deposits tokens to the bridge
- Parameters:
  - `tokenId` - Token identifier
  - `fromKey` - Sender's private key (or environment variable name)
  - `toKey` - Recipient's public key (or environment variable name)
  - `amount` - Amount to deposit (in whole units)
- Example: `protokit bridge deposit 1 SENDER_KEY RECIPIENT_KEY 100`

**`bridge redeem <tokenId> <toKey> <amount>`**
- Redeems tokens from the bridge
- Parameters:
  - `tokenId` - Token identifier
  - `toKey` - Recipient's public key (or environment variable name)
  - `amount` - Amount to redeem
- Example: `protokit bridge redeem 1 RECIPIENT_KEY 100`

**`bridge withdraw <tokenId> <senderKey> <amount>`**
- Withdraws tokens
- Parameters:
  - `tokenId` - Token identifier
  - `senderKey` - Sender's private key (or environment variable name)
  - `amount` - Amount to withdraw
- Example: `protokit bridge withdraw 1 SENDER_KEY 100`

### Settlement Deployment

**`settlement deploy`**
- Deploys settlement contracts
- Example: `protokit settlement deploy`

**`settlement token-deploy <tokenSymbol> <feepayerKey> <receiverPublicKey> [mintAmount]`**
- Deploys custom fungible tokens for settlement
- Parameters:
  - `tokenSymbol` - Symbol for the token (e.g., "USDC")
  - `feepayerKey` - Private key for paying deployment fees
  - `receiverPublicKey` - Public key to receive minted tokens
  - `[mintAmount]` - Amount to mint initially (default: 0)
- Example: `protokit settlement token-deploy 1 FEEPAYER_KEY RECEIVER_KEY 1000`

### Lightnet Utilities

**`lightnet wait`**
- Waits for the lightnet network to be ready before proceeding
- No positional parameters
- Example: `protokit lightnet wait`

**`lightnet faucet <publicKey>`**
- Sends MINA from the lightnet faucet to an account
- Parameters:
  - `publicKey` - Destination public key (or environment variable name)
- Example: `protokit lightnet faucet B62qnzbXQcUoQFnjvF4Kog6KfNsuuSoo7LSLvomPeak2CLvEYiUqT`

**`lightnet initialize`**
- Complete lightnet setup: waits for network, funds test accounts, deploys settlement
- No positional parameters
- Example: `protokit lightnet initialize`

### Developer Tools

**`generate-keys [count]`**
- Generates private/public key pairs for development
- Parameters:
  - `[count]` - Number of key pairs to generate (default: 1)
- Example: `protokit generate-keys 5`

**`explorer start`**
- Starts the explorer UI
- No positional parameters
- Options:
  - `-p, --port` - Port to run on (default: 5003)
  - `--indexer-url` - GraphQL endpoint URL for the indexer
- Example: `protokit explorer start -p 3000 --indexer-url http://localhost:8081/graphql`

All commands that interact with environment variable can be handled through two options:

### Option 1: `--env-path` (File-based)
Loads environment variables from a `.env` file
```bash
protokit settlement deploy --env-path ./my-config/.env
```

### Option 2: `--set KEY=value`
Passes environment variables directly as command-line arguments. Can be used multiple times.
```bash
protokit settlement deploy --set PROTOKIT_SETTLEMENT_CONTRACT_PRIVATE_KEY=EK... 
```

### Option 3: `--env {envName}`
Passes environment name.

```bash
protokit settlement deploy --env sovereign 
```

closes #353


packages/cli/src/index.ts
```diff
@@ -1,16 +1,21 @@
...
 import { hideBin } from "yargs/helpers";
 
-import { generateGqlDocsCommand } from "./commands/generateGqlDocs";
+import { parseEnvArgs } from "./utils/loadEnv";
 
 process.removeAllListeners("warning");
 process.env.NODE_NO_WARNINGS = "1";
 
 await yargs(hideBin(process.argv))
+  .scriptName("proto-kit")
```
rpanic: let's change this to `protokit`. I know I'd get really mad about that dash in the middle :)

packages/cli/src/index.ts
```diff
@@ -41,7 +49,349 @@ await yargs(hideBin(process.argv))
       }
     }
   )
-  .demandCommand()
-  .help()
+  .command(
```
rpanic: So - I don't really like the structure of this file - it's just a long spaghetti of commands.
We should find a pattern that allows this to be structured in a more modular way.
An idea would be that every file (command) implements a interface with a function `command<T, R>(argv: Argv<T>): Argv<R>` and internally calls `return argv.command(....)`
Then this file would only collect all those implementations and iterate over them

packages/cli/src/index.ts
```diff
@@ -41,7 +49,349 @@ await yargs(hideBin(process.argv))
...
+        await generateKeysCommand({ count: args.count });
+        process.exit(0);
+      } catch (error) {
+        console.error("Failed to generate keys:", error);
+        process.exit(1);
+      }
+    }
+  )
+  .command(
+    "lightnet:wait-for-network",
```
rpanic: So I had imagined this as a bit more reusable. Something like `protokit run deposit` where `run` says "execute a script" and `deposit` is the name of the script.
Where the script can either be one of our predefined scripts or a user-supplied one from the starter kit.
Now to realize this we have a few tools available that yargs gives us:
We can do nested commands: https://github.com/yargs/yargs/blob/main/example/nested.mjs
We can also let yargs go through the directories automatically and scaffold the command structure: https://github.com/yargs/yargs/blob/main/example/command_hierarchy.mjs
Some combination of this would be the most intuitive, create a mechanism to structure this on the framework side and then plug in the starter-kit side into the same mechanism

rpanic: More information on the folder stuff: https://github.com/yargs/yargs/blob/main/docs/advanced.md#providing-a-command-module

packages/cli/src/index.ts
```diff
@@ -41,7 +49,349 @@ await yargs(hideBin(process.argv))
...
+        });
+        process.exit(0);
+      } catch (error) {
+        console.error("Failed to wait for network:", error);
+        process.exit(1);
+      }
+    }
+  )
+  .command(
+    "lightnet:faucet <publicKey>",
```
rpanic: The namings like we talked

packages/cli/src/commands/lightnet/initialize.ts
```diff
@@ -0,0 +1,36 @@
+import { CommandModule } from "yargs";
+
+interface InitializeArgs {
+  "env-path"?: string;
+  env?: string[];
+}
+
+export const initializeCommand: CommandModule<{}, InitializeArgs> = {
+  command: "initialize",
```
rpanic, line 12: This on it's own is probably misleading, i.e. `protokit initialize` could mean anything.
So I'd group this, faucet and wait-for-network (wait) together into a `protokit lightnet` namespace.

rpanic (approved):

PR #416, opened 2026-02-01
enhance explorer ux

### LINKED ISSUE #389: Explorer improvements/follow-up
- Searching a block height shouldn't look for all the hashes that contain the block height, but instead I'd suggest searching for height only if the input is a number and less than the current max height (which we have access to, right?), and if the number is higher, we search for hashes

- Can we add the blocks view in the dashboard under the summaries? Kinda like https://etherscan.io/

- It would be cool if we could click on block hashes on the dashboard and get to the block's page right away

- For the transactions table inside the blocks view - can we use the same tick for the status that we use in the transaction detail page? Also, for failing transactions, lets display a x or something in red, while just adding the status message in the same column. I suspect most transactions to be successful, so the status message column would be mostly empty and it'd probably look better merged into one column.

- How hard is it to make the dashboard title and slogan configurable for users?

- Eventually, we should figure out if we want to display hashes as either hex or base64

- Also, we need to make a starter-kit PR to add the missing tasks to the module config

stanlou:
## Summary
- Follow-up PR to #381 addressing the review [comment](https://github.com/proto-kit/framework/pull/381#pullrequestreview-3643669629)

## Changes
- Search by height only when input is numeric and <= current max height; otherwise treat as hash.
- Add blocks view under dashboard  and link block hashes to block detail page.
- Update blocks tx table status: use the same success tick as tx detail page, show red “x” for failures, and merge status message into the same column.
- Make dashboard title and slogan configurable.

### DIFF
--- a/packages/explorer/src/components/search/GlobalSearch.tsx
+++ b/packages/explorer/src/components/search/GlobalSearch.tsx
@@ -1,5 +1,7 @@
 "use client";
 
+/* eslint-disable no-underscore-dangle */
+
 import { useCallback, useEffect, useState } from "react";
 import { useRouter } from "next/navigation";
 import { Search, Loader2, Zap, Cuboid, Link } from "lucide-react";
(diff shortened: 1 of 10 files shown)


packages/explorer/src/components/search/GlobalSearch.tsx
```diff
@@ -31,84 +33,158 @@ export interface SearchResponse {
...
+  const [maxHeight, setMaxHeight] = useState<number | null>(null);
   const router = useRouter();
 
-  const searchIndexer = useCallback(async (searchQuery: string) => {
-    if (!searchQuery || searchQuery.length < 2) {
-      setResults([]);
-      return;
-    }
+  useEffect(() => {
+    const fetchMaxHeight = async () => {
```
rpanic, line 55: How often is this updated?

packages/indexer/prisma/migrations/20260201142200_transaction_priority/migration.sql
```diff
@@ -0,0 +1,10 @@
+-- CreateTable
+CREATE TABLE "TransactionPriority" (
```
rpanic, line 2: Wait - do you have to also have this in the indexer? This is only used by the sequencer's mempool and nowhere else, and the foreign key structure should allow us to omit this in the indexer

rpanic (approved):

packages/explorer/src/components/search/GlobalSearch.tsx
stanlou: At the moment, it’s updated only on refresh. We could add a polling mechanism here, possibly extending it to other lists like blocks and transactions... ,does that make sense?

packages/indexer/prisma/migrations/20260201142200_transaction_priority/migration.sql
stanlou: it's used by [pushUserTransaction](https://github.com/proto-kit/framework/blob/f7fe5380abba5e6bd3e68b65b7f59de764c1a46e/packages/persistance/src/services/prisma/PrismaTransactionStorage.ts#L82) which is executed during transaction indexing.

PR #417, opened 2026-02-02
Made TimedBlockTrigger use independent timers

### LINKED ISSUE #403: Adjust TimedBlockTrigger to do different stages in parallel
These three things should happen in parallel, while we currently await completion of each stage.
- Block production
- Batch proving
- settlement & message rollup

They have to be coupled, since each stage's result is the input to the next stage, but especially batch proving and settlement contains a lot of waiting (for proofs to come and for transactions to be included). This shouldn't block the main thread and block production.
We already have the busy pattern which should implement the locks

rpanic:
Closes #403

### DIFF
--- a/packages/sequencer/src/mempool/private/PrivateMempool.ts
+++ b/packages/sequencer/src/mempool/private/PrivateMempool.ts
@@ -47,8 +47,12 @@ export class PrivateMempool
   }
 
   public async length(): Promise<number> {
-    const txs = await this.transactionStorage.getPendingUserTransactions(0);
-    return txs.length;
+    const numUserTxs =
+      await this.transactionStorage.countPendingUserTransactions();
+
+    const messages = await this.getMandatoryTxs();
+
+    return numUserTxs + messages.length;
   }
 
   public async add(tx: PendingTransaction): Promise<boolean> {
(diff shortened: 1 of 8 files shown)


4ytuncs (approved):

packages/sequencer/src/mempool/private/PrivateMempool.ts
```diff
@@ -47,8 +47,12 @@ export class PrivateMempool
   }
 
   public async length(): Promise<number> {
-    const txs = await this.transactionStorage.getPendingUserTransactions(0);
-    return txs.length;
+    const numUserTxs =
+      await this.transactionStorage.countPendingUserTransactions();
+
+    const messages = await this.getMandatoryTxs();
```
4ytuncs, line 53: Do you think if we also had a method like `getMandatoryTxCount` instead of  `getMandatoryTxs().length` would be beneficial?

4ytuncs (changes requested):

packages/sequencer/src/mempool/private/PrivateMempool.ts
rpanic: I thought about that, but the thing is that we have no real way of only evaluating the length. We always need to fetch, and internally, the IncomeMessagesService (on which this call is built upon) internally already caches the fetched messages, so the performance impact is very low - though it'd make the code nicer (but hide implementation nuances)

4ytuncs: It might be a subject of further code refactoring.

PR #469, opened 2026-03-23
Refactor/modules implementation

### LINKED ISSUE #458: Simplify a bunch of user-facing modules
- new `SequencerCoreModule` constructs
    - SequencerStartupModule
    - BlockProducerModule
    - BatchProducerModule -> Think about this, because inmemory doesn't technically need that
- PrismaRedisDatabase constructs:
    - DatabasePruneModule
- Rename LocalTaskWorker to WorkerModule
- Merge GraphqlServer and GraphqlServerModule into one (requires implementation of modulecontainer configs)
- BlockProducerModule
    - Move maximumBlockSize configuration to PrivateMempool and rename to more accurate targetBlockSize
    - Remove allowEmptyBlock config in favor of skipEmptyBlocks

stanlou:
closes #458


packages/sequencer/src/sequencer/SequencerCoreModule.ts
```diff
@@ -0,0 +1,133 @@
...
+}
+
+export interface SequencerCoreConfig extends InMemorySequencerCoreConfig {
+  BatchProducerModule: NoConfig;
+}
+
+@sequencerModule()
+@closeable()
+@dependencyFactory()
+export class InMemorySequencerCoreModule extends SequencerModule<InMemorySequencerCoreConfig> {
```
rpanic: We should change this naming, bcs this has nothing to do with InMemory itself.

packages/common/src/config/ModuleContainer.ts
```diff
@@ -155,10 +167,28 @@ export class ModuleContainer<Modules extends ModulesRecord>
...
+  protected readonly ownConfigKeys: Set<string> = new Set();
+
   public constructor(public definition: Modules) {
     super();
   }
 
+  /**
+   * Returns an object containing the keys listed in ownConfigKeys.
+   */
+  public get ownConfig(): OwnConfig {
```
rpanic: In general, I think this is good, a few notes:
Let's replace the ownConfigKeys mechanism by maing the OwnConfig type transform to `{ ownConfig: OwnConfig }` when passing it to ConfigurableModule (in the intersection type `CombinedModuleContainerConfig`) and then just accessing `this.config.ownConfig`

rpanic: Also let's rename ownConfig to containerConfig maybe? Or is that too misleading?

packages/api/src/graphql/GraphqlSequencerModule.ts
```diff
@@ -80,16 +132,91 @@ export class GraphqlSequencerModule<GraphQLModules extends GraphqlModulesRecord>
...
-          this.graphqlServer.registerSchema(module.generateSchema());
+          this.schemas.push(module.generateSchema());
         }
       }
     }
-    await this.graphqlServer.startServer();
+    await this.startServer();
+  }
+
+  // Server logic
```
rpanic, line 137: Just a thought - instead of flattening the Graphql implementation into here. Could we not just make this class a dependencyfactory creating the GraphqlServer? This way users could override it by passing a GraphqlServer module if they don't want to use Koa for some reason

packages/sequencer/src/sequencer/SequencerCoreModule.ts
```diff
@@ -0,0 +1,133 @@
...
+      },
+      blockProducerModule: {
+        useClass: BlockProducerModule,
+      },
+    };
+  }
+
+  public async start(): Promise<void> {
+    this.sequencerStartupModule.config = this.config.SequencerStartupModule;
+    this.blockProducerModule.config = this.config.BlockProducerModule;
```
rpanic, line 80: This works? I would've assumed that we should do this in create()...

packages/sequencer/src/sequencer/SequencerCoreModule.ts
```diff
@@ -0,0 +1,133 @@
...
+    private readonly sequencerStartupModule: SequencerStartupModule,
+    @inject("BlockProducerModule")
+    private readonly blockProducerModule: BlockProducerModule
+  ) {
+    super();
+  }
+
+  public static dependencies(): InMemorySequencerCoreDependencies {
+    return {
+      sequencerStartupModule: {
```
rpanic: What's up with the lowercase here? We normally do Uppercase for these in the codebase

packages/sequencer/src/sequencer/SequencerCoreModule.ts
```diff
@@ -0,0 +1,133 @@
...
+    this.blockProducerModule.config = this.config.BlockProducerModule;
+    this.batchProducerModule.config = this.config.BatchProducerModule;
+
+    await this.sequencerStartupModule.start();
+    await this.blockProducerModule.start();
+    await this.batchProducerModule.start();
+  }
+
+  public async close(): Promise<void> {
+    await this.sequencerStartupModule.close();
```
rpanic: Doubt that you need to do that - modulecontainer should pick up that sequencerStartupModule is Closeable and automatically call close() on it

packages/sequencer/src/storage/DatabasePruneModule.ts
```diff
@@ -14,7 +13,7 @@ export type DatabasePruneConfig = {
 
 @sequencerModule()
```
rpanic: Remove this annotation + Sequencermodule Superclass

rpanic (changes requested):

rpanic (approved):

rpanic (approved):

PR #474, opened 2026-03-25
More adaptive block result recomputation

rpanic:
Based on #473 

This PR adds a bit of stability towards missing block results. Specifically, it checks for missing results for every block to account for 1. failed result computation itself and more likely 2. network errors during database writes of the result


packages/sequencer/src/protocol/production/sequencing/BlockProducerModule.ts
```diff
@@ -239,19 +242,23 @@ export class BlockProducerModule extends SequencerModule<BlockConfig> {
     return blockResult?.block;
   }
 
-  public async blockResultCompleteCheck() {
+  public async blockResultCompleteCheck(): Promise<
```
stanlou, line 245: Naming feels a bit misleading since this also generates metadata, maybe `ensureLatestBlockResult`?

stanlou (approved):

PR #479, opened 2026-03-27
Integration/devnet module merge

rpanic:
Based on #469 and #475


packages/indexer/prisma/schema.prisma
```diff
@@ -90,8 +90,6 @@ model Block {
   beforeBlockStateTransitions Json @db.Json
 
   parentHash String? @unique
-  parent     Block?  @relation("Parent", fields: [parentHash], references: [hash])
```
stanlou, line 93: It looks like this relationship is still being used in [BlockFetching.ts](https://github.com/proto-kit/framework/blob/87c63ad626699d2ad58a83f8365794baf8085989/packages/processor/src/indexer/BlockFetching.ts#L66)

packages/indexer/src/tasks/IndexBlockTask.ts
```diff
@@ -24,26 +26,55 @@ export class IndexBlockTask
...
+    const missingTxs = results
+      .filter(([, result]) => result === undefined)
+      .map(([tx]) => tx);
+
+    const pushResults = await Promise.all(
+      missingTxs.map(
+        async (tx) => await this.transactionStorage.pushUserTransaction(tx, 0)
+      )
+    );
+    if (pushResults.some((x) => x)) {
```
stanlou: I'm not sure I understand the intent behind this check!

rpanic: Yeah that's a typo i think - should be `!x`

stanlou (approved):

PR #488, opened 2026-04-02
feat(indexer): recover missing blocks

### LINKED ISSUE #480: Indexer: Recovery missing blocks
Sometimes the indexer misses a block, which leads to incomplete data since the next block's parent won't exist in the DB. We should notice this happening and ask the sequencer for the missing block

stanlou:
closes #480


packages/indexer/src/tasks/IndexBlockTask.ts
```diff
@@ -12,17 +13,25 @@ import {
   IndexBlockTaskParametersSerializer,
 } from "./IndexBlockTaskParameters";
 
+export interface IndexBlockResult {
```
rpanic: So, I think for this we can type it as a union type
`{ status: "ok" } | { status: "missing-blocks", missingHeights: .... }` if we want.
But I leave it to you if we want to do it that way

packages/indexer/src/IndexerNotifier.ts
```diff
@@ -39,6 +45,67 @@ export class IndexerNotifier extends SequencerModule<Record<never, never>> {
...
+      const serialized = await this.indexMissingBlocksTask
+        .inputSerializer()
+        .toJSON(filteredBlocks);
+
+      await this.pushTask(
+        this.indexMissingBlocksTask.name,
+        this.indexMissingBlocksTask.name,
+        serialized
+      );
+    } catch (error) {
```
rpanic, line 99: Let's leave that to the queue implementation to handle errors that come from this. Generally, catching stuff early and not handling it (just emitting it is not handling it imo) lead to weird and non-recoverable behaviour. 
Throwing stuff is okay, sometimes stuff just breaks down, but then the sequencer should crash and restart.
At least that's the ideal philosophy I try to approach these things nowadays. Idk I just thought I'd share this - here it's definitely not missing critical so it might be unwise to let the sequencer crash because of a peripherial system. Yeah idk let's talk about it more i guess

packages/indexer/src/IndexerNotifier.ts
```diff
@@ -39,6 +45,67 @@ export class IndexerNotifier extends SequencerModule<Record<never, never>> {
...
+      }
+
+      const heights = [...result.missingHeights, result.incomingHeight];
+
+      const blocks = await Promise.all(
+        heights.map((h) => this.blockStorage.getBlockWithResultAt(h))
+      );
+
+      const filteredBlocks = blocks.filter(
+        (block): block is BlockWithResult => block !== undefined
```
rpanic: there is `filterNonUndefined` in the common package that we can reuse for this

packages/indexer/src/tasks/IndexBlockTask.ts
```diff
@@ -40,33 +41,48 @@ export class IndexBlockTask
...
+    input: IndexBlockTaskParameters[]
   ): Promise<IndexBlockResult> {
-    const incomingHeight = Number(input.block.height.toBigInt());
+    const firstBlockHeight = Number(input[0].block.height.toBigInt());
+
     try {
       const currentHeight = await this.blockRepository.getCurrentBlockHeight();
 
-      if (incomingHeight > currentHeight) {
+      if (firstBlockHeight > currentHeight) {
```
rpanic, line 92: Ideally, I'd like a few comments here on **why** this works (internal consistency given by the DB) bcs that fact is not immediately apparent

packages/indexer/src/tasks/IndexBlockTask.ts
```diff
@@ -40,33 +41,48 @@ export class IndexBlockTask
   public async prepare(): Promise<void> {}
 
   public async compute(
-    input: IndexBlockTaskParameters
+    input: IndexBlockTaskParameters[]
   ): Promise<IndexBlockResult> {
-    const incomingHeight = Number(input.block.height.toBigInt());
+    const firstBlockHeight = Number(input[0].block.height.toBigInt());
```
rpanic, line 83: Here - do we have the gurantee that those are in-order?

stanlou: We have two scenarios here:
- In normal indexing, we only receive a single block.
- If we receive multiple blocks, it means we’re indexing missing blocks that were generated using `Array.from()`

rpanic (approved):

PR #502, opened 2026-04-14
feat: add timestamp

stanlou:
Related to [10](https://github.com/proto-kit/explorer/pull/10)


CHANGELOG.md
```diff
@@ -7,7 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ## Unreleased
 
 ### Added
-
+- Added createdAt timestamp to transaction, block, batch and settlement models.[#502](https://github.com/proto-kit/framework/pull/502)
```
rpanic: In the future - let's add these to the bottom of the list s.t. PR numbers are ascending

packages/sequencer/src/mempool/PendingTransaction.ts
```diff
@@ -130,6 +130,7 @@ interface PendingTransactionJSONType {
     s: string;
   };
   isMessage: boolean;
+  createdAt?: Date;
```
rpanic: For Block, Batch and Settlement it makes sense to have a date - they are created at a specific point in time and that time has "meaning". Transactions only get their meaning by being included in a block and imo this should be the date we should in the explorer too.
Additionally, this date here can be freely set by the user the way it's implemented currently making it a bit though to rely on.

rpanic: So my suggestion would be to remove the createdAt from the transaction and only do it for block, batch, settlement. Then in the explorer we query the transaction's block and use that date

packages/sequencer/src/storage/model/Batch.ts
```diff
@@ -5,6 +5,7 @@ export interface Batch {
   proof: JsonProof;
   blockHashes: string[];
   height: number;
+  createdAt: Date;
```
rpanic: What do we think about making this a number and using `Date.now()`?
This would make serialization easier as it will not break the "json-serializable by default" property of these objects

rpanic (changes requested):

rpanic (approved):

PR #503, opened 2026-04-14
feat(cli): add scripts.env file support

### LINKED ISSUE #459: Implement CLI support for script-specific files
Import scripts.env from selected environment if they exist. 
CLI will look for that file variant first and only as a fallback use .env

Test this within the starter-kit and the (already existing scripts.env inside the sovereign env)

stanlou:
closes #459


packages/cli/src/scripts/settlement/deploy-token.ts
```diff
@@ -61,8 +67,8 @@ export default async function (
         ...protocol.settlementModules,
       }),
       Sequencer: Sequencer.from({
-        ...DefaultModules.prismaRedisDatabase(),
-        ...DefaultModules.settlementScript(),
+        Database: PrismaRedisDatabase,
```
rpanic: We shouldn't need a Database for deploy...

rpanic (changes requested):

